PR #15, opened 2016-10-05
IntroPython2016: add student dir jhefner and file placeholder

jmhthethird:


students/jhefner/README
```diff
@@ -0,0 +1,4 @@
+Whaddduuuuup
+This is a good ole fashioned README.
+With a few lines cuz it makes it easier to comment on stuff in a PR if there are actual comments.
```
codedragon: Thanks!

students/jhefner/session02/grid_printer.py
```diff
@@ -0,0 +1,8 @@
+"""
+grid printer
+"""
+
+def cheat_grid():
+    print("+ - - - - + - - - - +\n|         |         |\n|         |         |\n|         |         |\n|         |         |\n+ - - - - + - - - - +\n|         |         |\n|         |         |\n|         |         |\n|         |         |\n+ - - - - + - - - - +")
+
```
codedragon: Does the trick, but makes for a very long line, however, which is generally frowned upon. Here's a hint:
In [24]: '+' * 10
Out[24]: '++++++++++'

PR #19, opened 2016-10-05
Initial folder creation

heycasey:
Creating the initial folder

### DIFF
--- a/students/pcasey/session02/grid_printer.py
+++ b/students/pcasey/session02/grid_printer.py
@@ -1,0 +1,18 @@
+# Grid printer
+
+def print_grid():
+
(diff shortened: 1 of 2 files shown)


students/pcasey/session02/grid_printer.py
```diff
@@ -0,0 +1,18 @@
...
+
+    for i in range(2):
+        print(horiz)
+        for i in range(4):
+            print(mid)
+            i += 1
+        i += 1
+    print(horiz)
+
+print_grid()
```
codedragon: Larger grid than actually required, but I assume you can see how to make it any size required.

PR #25, opened 2016-10-05
uploading my grid lab

komashu:


codedragon:
Great start. Fun to see the solutions that avoid using loops. Now have to think about how to change the number of boxes.

komashu:
yeah i didn't get that far and didn't realize it was considered part of the homework!

PR #26, opened 2016-10-05
Session 2 Exercises

weidnem:
Not fancy. Git 'er done.


codedragon:
So, grid.py is empty. Was there code that you wanted us to look at?

weidnem:
Updated with actual grid exercise and fib exercise.

students/weidnem/session2/grid.py
```diff
@@ -0,0 +1,29 @@
...
+            pos += 1
+    else:
+        print()
+
+
+
+printgrid()
+
+
+
```
codedragon: This is a good start. Keep in mind that you can multiply strings. How would you change this to make the grid the size the user requests?

students/weidnem/session2/series.py
```diff
@@ -0,0 +1,23 @@
...
+def get_next(series,n):
+    if len(series) != n:
+        series.append(series[-2]+series[-1])
+        get_next(series,n)
+        
+    else:
+        print(series)
+
+
+fibonacci(9)
```
codedragon: This is a great. Note that the exercise was to return just the nth number in the series. Do you see how you can adapt this code to do the lucas exercise?

PR #27, opened 2016-10-05
jhefner: Session02

jmhthethird:


jmhthethird:
Updated with additional function so you can have any size of grid you wish.

students/jhefner/session02/series.py
```diff
@@ -0,0 +1,37 @@
...
+
+def fibonacci(n):
+    """This function returns the nth value in the fibonacci series,
+     derived from the functions input variable."""
+    if n<=1:
+        print(n)
+    else:
+        fib = (n-2)+(n-1)
+        print(fib)
+
```
codedragon: Do you still have questions about how to create fibonacci and lucas series? Feel free to ask if you don't know how to change your code. The main key is that you need to calculate the entire series up to n, but you are attempting to figure out just the nth number of the series with using only the last two numbers, and this is not possible. You can calculate the series with either a loop or recursion.

students/jhefner/session03/slicing_lab.py
```diff
@@ -0,0 +1,36 @@
...
+return a sequence with the first and last 4 items removed, and every other item in between
+return a sequence reversed (just with slicing)
+return a sequence with the middle third, then last third, then the first third in the new order
+NOTE: these should work with ANY sequence – but you can use strings to test, if you like.
+"""
+
+sequence = "This is definitely a sequence."
+
+first_last_exchanged = sequence[-1]+sequence[1:-1]+sequence[0]
+print(first_last_exchanged)
```
codedragon: sequence[-1] gives you a single object from your sequence. This works just fine for strings, but if you were using a list of numbers (sequence = [1, 2, 3, 4, 5], for example) it will pull that single number, and then you cannot add it to a list. To keep the single object as a sequence, use sequence[-1:]. for the first of the sequence, you need to do sequence[:1] rather than sequence[0], in order to keep it as a list of one item.
`sequence[-1:] + sequence[1:-1] + sequence[:1]`

students/jhefner/session03/slicing_lab.py
```diff
@@ -0,0 +1,36 @@
...
+
+every_other_item_removed = sequence[::2]
+print(every_other_item_removed)
+
+# Award for worst variable name goes to.....
+first_and_last_4_removed_every_other_item_between = sequence[4]+sequence[4:-5:2]+sequence[-5]
+print(first_and_last_4_removed_every_other_item_between)
+
```
codedragon: it's a bit long, but I wouldn't call it the worst. ;-) The worst is when the variable name has absolutely nothing to do with the object it is representing. This happens way too often, and I'd rather see long names than this.
You can do this one with a single slice. Your center slice is very close. (In fact, you are currently repeating the 5th element, but it is hard to see, since in your sequence the 5th element is a space). Play around with splicing with a list of numbers so you can see what it is doing more clearly.

students/jhefner/session03/slicing_lab.py
```diff
@@ -0,0 +1,36 @@
...
+sequence_reversed = sequence[::-1]
+print(sequence_reversed)
+
+
+third_of_a_sequence = int(len(sequence)/3)
+sequence_thirds_new_order = sequence[third_of_a_sequence*1:third_of_a_sequence*2]+ sequence[third_of_a_sequence*2:third_of_a_sequence*3]+sequence[third_of_a_sequence*0:third_of_a_sequence*1]
+print(sequence_thirds_new_order)
```
codedragon: Nice.

PR #32, opened 2016-10-06
Session 1 and Session 2 Work

spencermcghin:
Hi there,

Please check fibonacci and lucas number functions, as well as session 2 grid print.


spencermcghin:
Updated pull request with exercises through 30.

PR #33, opened 2016-10-06
fib series, final grid, fizzBuzz

Baumelbi:
This should be working versions of all of the labs


Baumelbi:
finished up the labs

students/baumel/session_02/fizzBuzz.py
```diff
@@ -0,0 +1,22 @@
+def fizzbuzz():
+    """ Write a program that prints the numbers from 1 to 100 inclusive.
+        But for multiples of three print “Fizz” instead of the number
+        For the multiples of five print “Buzz”.
+        For numbers which are multiples of both three and five print “FizzBuzz” instead"""
+
+    for x in range(1, 101):
+        if x % 3 == 0 and x % 5 == 0:
+            print('FizzBuzz')
+            x += 1
```
PythonCHB: no need to increment x -- it is getting reset by the for loop anyway.

students/baumel/session_02/grid_printer.py
```diff
@@ -11,4 +14,81 @@ def grid_printer():
...
 
-grid_printer()
+#version 2 using values as print objects
+def print_grid(n):
+    sides = "|"
+    apex = "+ "
+    topBottom = "- "
+    top = apex+topBottom*n+apex+topBottom*n+apex
+    wall = sides+"  "*(n)+" "
+    count = n
```
PythonCHB: is count used anywhere?

students/baumel/session_02/grid_printer.py
```diff
@@ -11,4 +14,81 @@ def grid_printer():
...
+    print(top)
+    wallMaker(n)
+    print(top)
+
+#top helper method
+def topMaker(n, squares):
+    apex = "+ "
+    lines = "- "
+    top = ((apex+lines*n)*squares)+apex
+    print(top)
```
PythonCHB: I like how you broke it out into functions -- but it would be a bit better to have these return the string, and then put all the printing in one place.

students/baumel/session_02/grid_printer.py
```diff
@@ -11,4 +14,81 @@ def grid_printer():
...
+    while counter != 0:
+        topMaker(size,squares)
+        wallMaker(size,squares)
+        counter -=1
+    topMaker(size, squares)
+
+
+def main():
+    squares = eval(input("How many squares would you like? "))
+    size = eval(input("what size squares would you like? "))
```
PythonCHB: careful with eval() -- it can be dangerous. You are expecting numbers, so you can use `int()`:

``squares = int(input("How many squares would you like? "))``

students/baumel/session_02/grid_printer.py
```diff
@@ -11,4 +14,81 @@ def grid_printer():
...
+        again(keepGoing)
+    else:
+        print_grid2(squares, size)
+        keepGoing = input("Would you like to make another grid (y/n)? ")
+        again(keepGoing)
+
+
+def again(keepGoing):
+    if keepGoing.lower() == 'y':
+        main()
```
PythonCHB: you want to be careful here -- this will work, but it's a recursive call -- it will keep piling up copies of the functions on the stack -- it left to run along time, it would overflow memory.

Better to use a while loop.

students/baumel/session_02/series.py
```diff
@@ -0,0 +1,16 @@
+def fibonacci(n):
+    """This program will return out the Nth value in the Fibonacci series"""
+    a, b = 1, 1
+    for i in range(n - 2):
+        a, b = b, a + b
```
PythonCHB: nice!

PR #35, opened 2016-10-07
session 2 labs (WIP)

aphollis:
could you provide an example of an appropriate use of Assert in my fibonacci exercise?


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PythonCHB: you don't generally want to add this kind of thing to git.

try removing it:

``git rm .idea/vcs.xml``

now that ``.idea`` is in ``.gitignore`` it shouldn't get added accidentally. 

But I generally avoid every using ``git add *`` or git add .``, but rather, add only the files you specifcally want to add,

students/aphollis/session02/fizzbuzz.py
```diff
@@ -0,0 +1,12 @@
+"""fizzbuzz"""
+
+for i in range(1,101):
+
+    if i % 3 == 0 and i % 7 == 0:
```
PythonCHB: it's supposed to be 3 and 5 -- but the logic is perfect.

students/aphollis/session02/series.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if len(list) == n:
+        return(list[n-1])
+    else:
+        list.append(next_up)
+        sum_series(n, list)
+
+
+fibonacci(6)
+lucas(6)
```
PythonCHB: So this where you can use asserts:

assert fibonacci(6) == 8

etc...

Ideally, you put them in an ``if __name__ == "__main__":`` block

```
if __name__ == "__main__":
    assert fibonacci(6) == 8
    etc...
```

PythonCHB:
I'm not going to merge until the .idea file is removed.

students/aphollis/session02/fizzbuzz.py
aphollis: fixed, no idea where 7 came from there...

aphollis:
removed .idea

PythonCHB:
nice job figuring out git!

PR #37, opened 2016-10-08
just finished grid printer exercise session 2

nachiketgalande:


students/liverpoolforever/session02/fizz_buzz.py
```diff
@@ -0,0 +1,18 @@
...
+"""
+
+# Declare the constants
+FIZZ = "FIZZ"
+BUZZ = "BUZZ"
+ITERATIONS = 21
+
+for i in range(1,ITERATIONS):
+    if i % 3 == 0 and i % 5 == 0 :
+        print(FIZZ,BUZZ)
```
PythonCHB: note that this will put a space between FIZZ and BUZZ -- which could be intended.

fine code.

students/liverpoolforever/session02/grid_printer.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+import sys
+print("=============== PRINT A GRID =================================")
+
+# Default no of rows and cols - corresponds to 3 rows and 3 cols
+# no_of_rows = 11
+# no_of_cols = 11
+
+# Get rows and cols from the user
+no_of_rows = sys.argv[1]
```
PythonCHB: nice job finding sys.argv!

students/liverpoolforever/session02/grid_printer.py
```diff
@@ -0,0 +1,60 @@
...
+VERTICAL_DASH = '|'
+BLANK_SPACE = " "
+SPACING = 5
+
+# returns the line with plus symbols
+def get_plus_line(cols,line):
+
+    for j in range(0,cols):
+        if j % SPACING == 0:
+            line = line + PLUS_SYMBOL
```
PythonCHB: it's a touch cleaner to use += here:

line += PLUS_SYMBOL

students/liverpoolforever/session02/series.py
```diff
@@ -0,0 +1,27 @@
...
+ITERATIONS = int(sys.argv[1])
+
+n1 = 0
+n2 = 1
+
+print("First 2 values in the series" , n1,n2)
+
+def fibonacci(n):
+    # Takes a parameter which the no of itertions to be computed
+    global n1,n2
```
PythonCHB: better to avoid globals. either pass n1 and n2 into teh function, or, better yet dimply call fibonacci twice recursively:

return fibonacci(n-1) + fibonacci(n-2)

students/liverpoolforever/session02/series.py
```diff
@@ -0,0 +1,27 @@
...
+    if( n > 0):
+        n3 = n1 + n2
+        n1 = n2
+        n2 = n3
+        print("Value: ", n3)
```
PythonCHB: also -- better for the function to compute an return the value, then you can call it if you want to print it.

PR #38, opened 2016-10-08
Readme File

psbriant:
Added Readme File


students/psbriant/session02/series.py
```diff
@@ -0,0 +1,81 @@
...
+def sum_series(n, first=0, second=1):
+    """
+    Return the nth value in either the Fibonacci series or the Lucas series
+    based on the values of integer arguments first and second.
+    """
+    if n == 0:
+        return first
+    if n == 1:
+        return second
+    return sum_series(n-1, first, second) + sum_series(n-2, first, second)
```
PythonCHB: perfect!

students/psbriant/session02/series.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+# ==============================================================================
+
+
+def main():
+    """
+    Tests each method to verify specified parameters return expected output.
+    """
+
```
PythonCHB: loving all the tests!

students/psbriant/session03/list_lab.py
```diff
@@ -0,0 +1,126 @@
...
+def series3(fruit):
+    """
+    Take in a list of strings and iterate through it prompting the user
+    regarding their preference for each fruit. If the user does not like the
+    fruit, it is deleted from the list. If the user does not specify their
+    preference, they are prompted until they do. A list of remaining fruit is
+    displayed after the user has specified their preference for each fruit.
+
+    Questions:
+    Best way to ensure each element is lowercase.
```
PythonCHB: the str.lower() method makes a string lower case.

I often do comparisons with input with something like:

if fruit[0].lower() == 'p'

then it will catch both cases.

students/psbriant/session03/list_lab.py
```diff
@@ -0,0 +1,126 @@
...
+    How to avoid skipping elements when the element before them is deleted.
```
PythonCHB: if you are going to be altering a list while looping though it (particularly deleting elements) then you want to loop though a copy:

```
for item in fruit[:]:
   ... whatever logic
   fruit.remove(item)
```
you are altering the fruit list, but the for loop is loping through a copy, so it will all work.

students/psbriant/session03/list_lab.py
```diff
@@ -0,0 +1,126 @@
...
+        # Asks the user whether they like each fruit
+        preference = input("Do you like " + item.lower() + "? yes|no ")
+        # Checks to make sure user entered either 'yes' or 'no'
+        if preference != 'no' and preference != 'yes':
+            # Repeatedly prompts user to enter 'yes' or 'no' until they do.
+            while True:
+                preference = input("Please answer with either yes or no ")
+                if preference == 'yes' or preference == 'no':
+                    # Ask about using break instead of False
+                    return False
```
PythonCHB: yes, break is a better option here.

students/psbriant/session03/list_lab.py
```diff
@@ -0,0 +1,126 @@
...
+    fruit, it is deleted from the list. If the user does not specify their
+    preference, they are prompted until they do. A list of remaining fruit is
+    displayed after the user has specified their preference for each fruit.
+
+    Questions:
+    Best way to ensure each element is lowercase.
+    How to avoid skipping elements when the element before them is deleted.
+    Best way of ending the while loop.
+    Correctness of using break and when not to use break.
+    """
```
PythonCHB: break is what you usally want to do to end a while loop.

A "while True" loop is infinite -- it had better have a break in in somewhere...

(though if the function is really done, you can use a return, but it's a bit klunky -- hard to see where the function ends at a glance.

students/psbriant/session03/list_lab.py
```diff
@@ -0,0 +1,126 @@
...
+
+def series4(fruit):
+    """
+    Take in a list of strings and reverse each letter of each item. Display the
+    original list minus the last element and it's copy.
+    """
+    # Create copy of fruit
+    rev_fruit = fruit[:]
+    rev_fruit = " ".join(rev_fruit)
+    rev_fruit = rev_fruit[::-1]
```
PythonCHB: slick! I didn't think of that -- I was thinking to loop through the list and reverse each item -- but this is pretty neat.

students/psbriant/session03/slicing.py
```diff
@@ -0,0 +1,66 @@
...
+    Hi
+    """
+    first_last('mississippi')
+    every_other('mississippi')
+    four_removed('mississippi')
+    reversed1('mississippi')
+    third_is_the_word('mississippi')
+
+
+if __name__ == '__main__':
```
PythonCHB: this all looks great.

PR #43, opened 2016-10-11
add series.py

heb2016:
Exercise Fibanacci

### DIFF
--- a/students/heb2016/session02/series.py
+++ b/students/heb2016/session02/series.py
@@ -1,0 +1,91 @@
+
+## Session02 Exercise 
+# fibonacci 
+def fibonacci(n):
+		row=0
+		while row<=n:
+			if row==0:
+				nth1=row
+				nth2=0
+				nth=nth1+nth2
+				if n==0:
+					print(row, "number", nth)
+					return(nth)
+				row+=1
+			if row==1:
+				nth2=row
+				nth1=0
+				nth=nth1+nth2
+				if n==1:
+					print(row,"number", nth)
+					return(nth)
+				row+=1
+			nth=nth1+nth2
+			if row>1 and row==n:
+				print(row, "number",nth)
+				return(nth)
+			row+=1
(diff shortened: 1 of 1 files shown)


students/heb2016/session02/series.py
```diff
@@ -0,0 +1,91 @@
+
+## Session02 Exercise 
+# fibonacci 
+def fibonacci(n):
+		row=0
+		while row<=n:
```
PythonCHB: makes more sense to use a for loop when you know when it will terminate:

```
for row in range(n):
```

students/heb2016/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+			if row==0:
+				nth1=row
+				nth2=0
+				nth=nth1+nth2
```
PythonCHB: this seems like more code than you need -- why not use "row" (or a better name), rather than assigning it to nth1?

you can also use Python's multiple assignment:

nth1, nth2 = row, 0

students/heb2016/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+				if n==0:
+					print(row, "number", nth)
+					return(nth)
+				row+=1
+			if row==1:
+				nth2=row
+				nth1=0
+				nth=nth1+nth2
```
PythonCHB: this is the same code in each if clause -- you could probably put that outside of the "if".

students/heb2016/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+#print( "Lucas")
+#lucas(0)
+#lucas(1)
+#lucas(2)
+#lucas(3)
+
+
+## Generalization
+def sum_series(n, n1, n2):
+	if n1 ==' ':
```
PythonCHB: and an single space is not a great default -- if a number isn't appropriate, use None for a undefined value.

PR #44, opened 2016-10-12
adding my three labs.

komashu:


students/shulatif/session_02/fizzbuzz.py
```diff
@@ -0,0 +1,20 @@
+
+
+""" Print every number 1 - 100. But for multiples of 3, print Fizz. For multiples of 5 print Buzz. And for multiples
+of 3 AND 5 print FizzBuzz.
+"""
+
+
+def fizzbuzz(n):
+    count = 0
+    while count <= n:
```
PythonCHB: here's where a for loop makes the most sense:

``for count in range(n+1):``

students/shulatif/session_02/grid_printer.py
```diff
@@ -3,11 +3,10 @@
...
 
 
-def grid(i):
-    cross = '+ ' + '- ' * i + '+ ' + '- ' * i + '+\n'
+def grid(i, x):
+    cross = '+ ' + '- ' * i + '+ ' + '- ' * i + '+\n' '| ' + '  ' * i + '| ' + '  ' * i + '|\n'
     blanks = '| ' + '  ' * i + '| ' + '  ' * i + '|\n'
     middle = blanks * i
-
     print(cross+middle+cross+middle+cross)
```
PythonCHB: this all looks great.

students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
+def fibonacci(n):
+    """
+        This will take two numbers in a series and add them together to get the next number, starting with 0 and 1.
+        We don't want this to go on for infinity. Using the if/else statements, since it starts at 0 and 1, I just
+        used the value of the counter to start off. Then it cycles through adding the numbers to each other and
+        raising the counter number to stop the while loop once it reaches the initial Fn.
+    """
+    count = 0
+    fib0, fib1 = 0, 1
+    while count < n:
```
PythonCHB: another good use for a for loop :-)

students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
...
+        if count <= 1:
+            fibo = count
+        else:
+            fibo = fib0 + fib1
+            fib0 = fib1
+            fib1 = fibo
```
PythonCHB: multiple assignment is helpful here:

fib0, fib1 = fib1, fibo

students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
...
+        count += 1
+
+
+def sum_series(y, n=0, x=1):
+    """
+        This requires one parameter and the other two are optional. The default will produce the fibonacci numbers.
+        But if 2 and 1 are used, it will produce the lucas numbers. If all the parameters are used, it will use the
+        second and third value as a starting point with the initial parameter for the iterations.
+    """
+    if not n and not x:
```
PythonCHB: this would be both zero -- is that what you want?

students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
...
+                randos = x
+            else:
+                randos = n + x
+                n = x
+                x = randos
+            print(randos)
+            count += 1
+
+
+class TestOfMathFunctions:
```
PythonCHB: getting fancy sith test classes? nice, but maybe wait 'till we get there in class.

but you can do some tests with simple assert statements:

assert fibonacci(5) == 5

etc....

komashu: yeah i wasn't really clear on the tests. it's definitely something i want to learn more about since they're so important.

PR #45, opened 2016-10-12
homework

bc-uw:


students/bc-uw/session02/fib.py
```diff
@@ -0,0 +1,16 @@
...
+    The Fibonacci sequence starts with 0 and 1 and adds them together.
+    From there, it adds the result to the previous two integers.
+    The sequence is 0, 1, 1, 2, 3, 5, 8, 13..
+    The formula is fib(n) = fib(n-2) + fib(n-1)
+    """
+    if n == 0:
+        return 0
+    if n == 1 or n == 2:
+        return 1
+    else: return Fibonacci(n-2) + Fibonacci(n-1)
```
PythonCHB: great solution.

put the return on the next line -- this works, but it's a bit more readable if the code block is more clear.

students/bc-uw/session02/fizzbuzz.py
```diff
@@ -0,0 +1,16 @@
...
+
+def fizzbuzz():
+    for i in range(1, 101):
+        if i % 3 == 0 and i % 5 == 0:
+            print("FizzBuzz")
+        elif i % 3 == 0:
+            print("Fizz")
+        elif i % 5 == 0:
+            print("Buzz")
+        else: print i
```
PythonCHB: put that print on the next line, but otherwise a fine solution.

students/bc-uw/session02/grid_printer.py
```diff
@@ -0,0 +1,42 @@
...
+|               |               |
+|               |               |
+|               |               |
++ - - - - - - - + - - - - - - - +
+
+"""
+
+def make_grid(size, boxes):
+    """
+    This is gross, but it works.
```
PythonCHB: not so gross -- and working code is good code!

PR #47, opened 2016-10-12
All work for the week!

jerbear4:


jerbear4:
Working on week 3 assignments.

PythonCHB:
whew! that's a lot!

I'll try to review as much as I can :-)

students/jbearer/session01/break_me.py
```diff
@@ -0,0 +1,14 @@
...
+# x = 5
+# y = "foo"
+# bar = x + y
+
+# Generates SyntaxError
+# 2foo = 100
+
+# Generates AtributeError
+# foo = "bar"
+# foo.count = 25
```
PythonCHB: interesting! you DO get an AttributeError, but NOT because the attribute is not there -- but rather because it's not settable:

`AttributeError: 'str' object attribute 'count' is read-only`

str.count is a valid method -- but it's a built-in and you can't set it.

try:

```
In [4]: str.couunt
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-33d30b811966> in <module>()
----> 1 str.couunt

AttributeError: type object 'str' has no attribute 'couunt'
```
different error message, because with the typo it doesn't exist at all.

students/jbearer/session02/booleans.py
```diff
@@ -0,0 +1,18 @@
+def near_ten(num):
+    return (num % 10) == 0 or (num % 10) == 1 or (num % 10) == 2 or (num % 10) == 8 or (num % 10) == 9
+
```
PythonCHB: how about:
`if  (0 <= (num % 10) <= 1) or (8 <= (num % 10) <= 9)`

for a little less reptition...

students/jbearer/session02/gridprinter.py
```diff
@@ -0,0 +1,66 @@
+# Grid Printer Exercise
+# Session02 - Oct. 5, 2016
+
+def simple_grid():
+    """ Prints a simple grid """
+    print("Simple Grid")
+    print("+----+----+")
+    print("|" + "    " + "|" + "    " + "|")
```
PythonCHB: this works fine -- but can you have less repition by using string multiplication?

or a for loop...

students/jbearer/session02/gridprinter.py
```diff
@@ -0,0 +1,66 @@
...
+    print("|" + "    " + "|" + "    " + "|")
+    print("+----+----+")
+    print()
+
+simple_grid()
+
+""" ================================================== """
+
+def print_dyn(n):
```
PythonCHB: there we go! nice.

students/jbearer/session02/lpthw_ex11.py
```diff
@@ -0,0 +1,8 @@
+print("How old are you? ", end="")
```
PythonCHB: I'm not going to review al this LPTHW, unless you have questions...

students/jbearer/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+        return 0
+    if n == 1:
+        return 1
+
+    f1 = 0
+    f2 = 1
+    
+    for i in range(0,n-1):
+        fn = f1 + f2
+        f1 = f2
```
PythonCHB: you can make this a tiny bit more compact with:

`f1, f2 = f2, fn`

or even:

```
f1, f2 = f2,  f1 + f2
return f2
```

students/jbearer/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+        f2 = fn
+
+    return fn
+
+# print("Fibonacci: " + str(fibonacci(3)))
+
+
+def lucas(n):
```
PythonCHB: these are fine iterative solutions -- but make sure you know how to di it recursively as well -- i.e. caling teh function from within itself....

students/jbearer/session02/series.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def sum_series(x, y=0, z=1):
+    """
+    Optional parameters determine whether Fibonacci or Lucas values will be returned.
+    If optional parameters are 0 and 1, return Fibonacci
+    If optional parameters are 2 and 1, return Lucas
+    """
+    if y == 0 and z == 1:
+        return fibonacci(x)
```
PythonCHB: this works fine, of course, but not for any other series if this form.

The idea was to make a general function that could compute these two -- or any other of teh same form, with two different starting values.

I know -- confusing, as you had just written the other two....

but you got the default parameters, which was the main goal here.

students/jbearer/session03/list_lab.py
```diff
@@ -0,0 +1,55 @@
...
+    
+    print()
+    print('-> List Lab: Section 2 <-')
+    print('Starting list:', l)
+
+    l.pop()
+    print('Last fruit removed:', l)
+
+    del_fruit = input('Which fruit would you like to delete? ')
+    l.pop(l.index(del_fruit))
```
PythonCHB: pop() is not ideal if you don't have an index:

`l.remove(del_fruit)`

students/jbearer/session03/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
...
+    l[0] = last
+    l[-1] = first
+
+    print(l)
+exch_func()
+
+
+def skip_list():
+    l = [ '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]
+    new_l = l[0:len(l):2]
```
PythonCHB: you almost never need to use len() in slicing:

`l[::2]` will do it.

students/jbearer/session03/slicing_lab.py
```diff
@@ -0,0 +1,38 @@
...
+def reverse_func():
+    l = [ '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]
+    new_l = l[::-1]
+    print(new_l)
+reverse_func()
+
+def middle3_last3():
+    l = [ '0', '1', '2', '3', '4', '5', '6', '7', '8', '9' ]
+    i = len(l) // 3
+    new_l = l[i:i+3] + l[-i:] + l[:i]
```
PythonCHB: this only works if the list happens to be len 9. why not:

`l[i:2*i] ....`

PR #49, opened 2016-10-14
List and String Lab

jerbear4:


students/jbearer/session03/list_lab.py
```diff
@@ -27,21 +27,29 @@ def fruit_tree():
...
-    print('-> List Lab: Section 2 <-')
+    print('-> List Lab: Section 2 and BONUS<-')
     print('Starting list:', l)
 
     l.pop()
     print('Last fruit removed:', l)
-
-    del_fruit = input('Which fruit would you like to delete? ')
-    l.pop(l.index(del_fruit))
+    l = l[:] * 2
```
PythonCHB: l * 2 will make a new list -- no need for the [:] copy.

students/jbearer/session03/list_lab.py
```diff
@@ -27,21 +27,29 @@ def fruit_tree():
...
+    print('Multiply string * 2:',l)
+    del_chk = True
+    while del_chk == True:
```
PythonCHB: you can jsut do: 

`while del_chk:`

no need for == True, True is already "truthy"

students/jbearer/session03/list_lab.py
```diff
@@ -27,21 +27,29 @@ def fruit_tree():
...
+        del_fruit = input('Which fruit would you like to delete? ')
+        if del_fruit in l:
+            while del_fruit in l:
+                l.remove(del_fruit)
+            del_chk = False
+        else:
```
PythonCHB: no need for:

```
else:
    continue
```
here -- continuing is the default anyway.

students/jbearer/session03/list_lab.py
```diff
@@ -27,21 +27,29 @@ def fruit_tree():
...
+            continue
     print('Current list:', l)
 
     print()
     print('-> List Lab: Section 3 <-')
     i = 0
     while i < len(l):
```
PythonCHB: this is where a for loop makes more sense -- you want to loop through the whole list:

```
for fruit in l[:]: # making a copy so I can delete from the original
    answer = input('Do you like %s (yes / no)? ' % l[i].lower())
    ...
```

students/jbearer/session03/rot13.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+"""
+Excercise --  ROT13
+October 13, 2016
+"""
+
+if __name__ == '__main__': # <- Would like a better idea of what exactly this does
```
PythonCHB: I'll talk more about that next class -- I'm sure you're not the only one.

students/jbearer/session03/rot13.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def rot13(myStr):
+        alpha = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'
+        key =   'nopqrstuvwxyzabcdefghijklmNOPQRSTUVWXYZABCDEFGHIJKLM'
+        decoder = str.maketrans(alpha,key)
+        answer = myStr.translate(decoder)
```
PythonCHB: nice job finding str.maketrans and str..translate()!

very nice answer.

students/jbearer/session03/string_formatting.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+"""
+String Formatting Lab -- October 13, 2016
+"""
+import random
+
+print('file_00{:} :     {:.2f},  {:.0e}'.format(2,123.4567,10000))
```
PythonCHB: the idea here is that you might have file numbers anywhere from 1 to 999, but you want them zero padded so they will sort correctly. So this should work with a range of numbers, not only with '2' or other single digits. 

the other ones are perfect.

students/jbearer/session03/string_formatting.py
```diff
@@ -0,0 +1,13 @@
...
+
+x = random.randrange(101)
+y = random.randrange(101)
+z = random.randrange(101)
+
+print('The first 3 numbers are: {:d}, {:d}, {:d}'.format(x, y, z))
```
PythonCHB: I didn't really expect this to all get done, as we didn't get to it in class...

So I'll merge this for now.

PR #50, opened 2016-10-15
Finally finished print_report. Had a heck of a time trying to figure out how to sort list...

jerbear4:
…out how to sort.

### DIFF
--- a/students/jbearer/session03/mailroom.py
+++ b/students/jbearer/session03/mailroom.py
@@ -8,15 +8,15 @@ donor_list = {'Shrek':[5,10], 'Donkey':[23,101,4], 'Princess Fiona':[7,28],
 'Puss in Boots':[36,12,10], 'Gingerbread Man':[12,34]}
 
 def print_report():
-    print('### Donor Report ###')
-    print('{:16} {:6} {:6}'.format('Name','Total','Donations'))
+    print('{:^33}'.format('### Donor Report ###'))
+    print('{:25} {:10} {:10}'.format('Name','Total','Donations'))
     sort_l = []
     for name, donations in donor_list.items():
         sort_l.append((name, sum(donations), len(donations)))
     
     sort_l = sorted(sort_l, key=lambda donations: donations[1], reverse=True)
     for i in sort_l:
-        print('{:15} {:6} {:6}'.format(i[0],i[1],i[2]))
+        print('{:20} {:10} {:10}'.format(i[0],i[1],i[2]))
 
 def proc_donation(d_name):
     amt = True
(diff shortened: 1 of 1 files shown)


students/jbearer/session03/mailroom.py
```diff
@@ -1,9 +1,81 @@
...
 """
 
-if __name__ == '__main__':
+donor_list = {'Shrek':[5,10], 'Donkey':[23,101,4], 'Princess Fiona':[7,28],
+'Puss in Boots':[36,12,10], 'Gingerbread Man':[12,34]}
 
-    
\ No newline at end of file
+def print_report():
+    print('{:^33}'.format('### Donor Report ###'))
```
PythonCHB: no need for formatting here, a simple:

print(''### Donor Report ###')

would be fine.

students/jbearer/session03/mailroom.py
```diff
@@ -1,9 +1,81 @@
...
+    print('{:25} {:10} {:10}'.format('Name','Total','Donations'))
+    sort_l = []
+    for name, donations in donor_list.items():
+        sort_l.append((name, sum(donations), len(donations)))
+    
+    sort_l = sorted(sort_l, key=lambda donations: donations[1], reverse=True)
```
PythonCHB: very nicely done! and you even found lambda!

students/jbearer/session03/mailroom.py
```diff
@@ -1,9 +1,81 @@
...
+    for i in sort_l:
+        print('{:20} {:10} {:10}'.format(i[0],i[1],i[2]))
+
+def proc_donation(d_name):
+    amt = True
+    while amt == True:
+        try:
+            donation = int(input('Enter donation amount: '))
+        except ValueError:
```
PythonCHB: awesome! just the right way to handle this.

students/jbearer/session03/mailroom.py
```diff
@@ -1,9 +1,81 @@
...
+    return
+
+
+def send_thanks():
+    chk = True
+    while chk == True:
+        name = input('Enter name or type "list": ')
+        if name == 'list':
+            for d_names in donor_list.keys():
+                print(d_names)
```
PythonCHB: probably better to have a separate function to do the printing, rather than having the interactive loop have to know about the data structures you're using.

students/jbearer/session03/mailroom.py
```diff
@@ -1,9 +1,81 @@
...
+            print_report()
+        elif response == 's':
+            send_thanks()
+        elif response == 'x':
+            break
+        else:
+            print('please type "s", "p", or "x"')
+
+if __name__ == "__main__":
+    main()
```
PythonCHB: nice job -- looking good!

students/jbearer/session03/mailroom.py
jerbear4: When we go over lambda in class, I'll have some questions here.  I found this example code on the python.org website.  Still trying to wrap my head around how it actually works.

students/jbearer/session03/mailroom.py
jerbear4: I can't take credit for the beginning menu system.  It was all part of the example that was provided in the examples folder.  I just added the functionality for the rest of the program.

PR #51, opened 2016-10-15
Week3

BenBagBag:
Donor report would have made so much more sense with dictionaries... guess that's next week, huh? :)


PythonCHB:
Exactly -- next week's homework is to update this with dicts :-)

students/cowhey/session03/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+print(fruits)
+
+# Ask the user for a number
+# display the number back to the user and the fruit corresponding to that number (on a 1-is-first basis).
+fruit_index = input("Give me a number of a fruit: ")
+
+if int(fruit_index) < len(fruits):
+    print(fruits[(int(fruit_index) - 1)])
+else:
+    print("Not a fruit. Have a marshmallow.")
```
PythonCHB: nice!

students/cowhey/session03/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# Ask the user for input displaying a line like “Do you like apples?”
+# for each fruit in the list (making the fruit all lowercase).
+fruits_copy = fruits[:]
+for fruit in fruits_copy:
+    like = input("Do you like {}?".format(fruit.lower()))
+    acceptable_answers = ["yes", "y", "no", "n"]
+    # For any answer that is not “yes” or “no”, prompt the user to answer with one of those two values (a while loop is good here):
+    while like.lower() not in acceptable_answers:
```
PythonCHB: very nice.

students/cowhey/session03/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+    # For each “no”, delete that fruit from the list.
+    if like.lower() in acceptable_answers[2:]:
+        fruits.remove(fruit)
+
+# Display the list.
+print(fruits)
+
+# Make a copy of the list and reverse the letters in each fruit in the copy.
+# use copy from above
+reversed_fruits = [fruit[::-1] for fruit in fruits_copy]
```
PythonCHB: list comprehension! very nice.

students/cowhey/session03/list_lab.py
```diff
@@ -0,0 +1,79 @@
...
+print(reversed_fruits)
+
+# Delete the last item of the original list. Display the original list and the copy.
+del fruits[-1]
+print(fruits)
+print(fruits_copy)
```
PythonCHB: This all looks great.

students/cowhey/session03/mailroom.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
+
+def accept_donation():
+    amount = input("How much did this person donate? \n")
+    if amount == "x" or check_if_number(amount):
+        return amount
```
PythonCHB: you may want to return a numerical representation of amount here, rather than the string that input() returned

students/cowhey/session03/mailroom.py
```diff
@@ -0,0 +1,135 @@
...
+def create_reports():
+    donor_rows = []
+    for donor in donors:
+        new_row = []
+        new_row.append(donor[0])
+        new_row.append("{:.2f}".format(sum(donor[1:])))
+        new_row.append(str(len(donor[1:])))
+        new_row.append("{:.2f}".format(float(new_row[1]) / int(new_row[2])))
+        donor_rows.append(new_row)
+    return donor_rows
```
PythonCHB: nice -- I like how you returned the list to print, rather than printing it here.

students/cowhey/session03/mailroom.py
```diff
@@ -0,0 +1,135 @@
...
+# here is where triple quoted strings can be helpful
+msg = """
+What would you like to do?
+
+To send a thank you: type "s"
+To print a report: type "p"
+To exit: type "x"
+"""
+
+donors = [["Wilmot Filipe", 18.00, 72.00], ["Neoptolemus Yaropolk", 36.00], ["Mahesha Diogenes", 90.00, 54.00, 18.00], ["Arthur Tjaz", 18.00], ["Luuk Sinclair", 180.00]]
```
PythonCHB: here is can be helpful to add another level of nesting:

`` [["Wilmot Filipe", [18.00, 72.00]], ...``

note the extra nested list for the donations -- it makes it easier to extract those.

students/cowhey/session03/questions.txt
```diff
@@ -0,0 +1,4 @@
+questions:
+-why can't i make a new list or tuple using a list comprehension-like function on a tuple?
+(given that a tuple is iterable)
```
PythonCHB: you can :-)

[item for item in a_list]

will make a copy.

but: 

1) we haven't learned list comps yet 
2) it's a bit more overhead, and more code than simply:
a_list[:]
3) if you do that with a tuple, you'll get a list back -- there is not "tuple comprehension"

students/cowhey/session03/questions.txt
```diff
@@ -0,0 +1,4 @@
...
+-what is reversed(x) 's resultant object? what good is it?
```
PythonCHB: try it:

```
In [14]: type(reversed(l))
Out[14]: list_reverseiterator
```

so it's a "list_reverseiterator" object -- not helpful, is it :-)

but what it means is that it's an "iterable" that you can then do things like loop through with a for loop, etc. but it hasn't made a copy of the list -- it returns the items one by one as they are asked for. this has performance benefits, as it doesn't have to make a copy of the whole thing.

so you use it if you want to loop through something in reversed order, but dont actually need an actual list with the order reversed.

students/cowhey/session03/rot13.py
```diff
@@ -0,0 +1,43 @@
+def rot13(encrypted_str):
+    '''Return an unencrypted string that was encrypted with ROT13.
```
PythonCHB: did yo notice that rot13 is reversible? it's the same operation to go from encrypted to unencrypted as the other way around...

students/cowhey/session03/rot13.py
```diff
@@ -0,0 +1,43 @@
...
+    Uses the str.maketrans() method'''
```
PythonCHB: nice find!

students/cowhey/session03/rot13.py
```diff
@@ -0,0 +1,43 @@
...
+    '''Return an unencrypted string that was encrypted with ROT13.
+    Does not use the str.maketrans() method'''
+    decoded_chars = []
+    for char in encrypted_str:
+        decoded_chars.append(shift_letter(char))
+    return ''.join(decoded_chars)
+
+
+if __name__ == '__main__':
+    assert(rot13("Zntargvp sebz bhgfvqr arne pbeare") == "Magnetic from outside near corner")
```
PythonCHB: nice tests!

students/cowhey/session03/slicing.py
```diff
@@ -0,0 +1,32 @@
...
+
+def reverse(sequence):
+    '''return a sequence reversed (just with slicing)'''
+    return sequence[::-1]
+
+
+def mix_thirds(sequence):
+    '''return a sequence with the middle third, then last third, then the first third in the new order'''
+    third = len(sequence) // 3
+    return sequence[third:2*third] + sequence[2*third:] + sequence[0:third]
```
PythonCHB: all good.

students/cowhey/session03/upriss_list.py
```diff
@@ -0,0 +1,64 @@
...
+    print("Pay more attention next time, that student doesn't exist.")
+# Add "John Smith" and "Mary Miller" at the beginning of the list (by using "+")
+student_list = ["John Smith", "Mary Miller"] + student_list
+print(student_list)
+# Remove the last name from the list.
+student_list.pop()
+print(student_list)
+# Ask a user to type a name.
+remove_student = input("Give me a name on the list.")
+standardized_names = [name.lower() for name in student_list]
```
PythonCHB: nice! -- jumping ahead to comprehensions.

PR #52, opened 2016-10-16
rot13.py commit

johicks:


students/josh_hicks/session03/rot13.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+# ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz
+# NOPQRSTUVWXYZABCDEFGHIJKLMnopqrstuvwxyzabcdefghijklm
+
+
+def rot13(s):
+    # Is there an easier way to construct this dictionary?
```
PythonCHB: yup :-) a few options. the obvious one is something like:

```
for first, second in zip(original, translated):
    rot13[first] = second
```

or even easier:

```
rot13 = dict(zip(original, translated)
```

or even easier, from the docs:

static str.maketrans(x[, y[, z]])

...
If there are two arguments, they must be strings of equal length, and in the resulting dictionary, each character in x will be mapped to the character at the same position in y.

students/josh_hicks/session03/slicing.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+l = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+
+
+def firstlast_exchange(l):
+    flipped_ends = l[1:-1]
+    flipped_ends.append(l[0])
+    flipped_ends.insert(0, l[-1])
+    return flipped_ends
```
PythonCHB: this will only work with mutable sequences -- i.e. lists. you can do it with plain old slicing:

`return l[-1] + l[1:-1] + l[0]`

students/josh_hicks/session03/slicing.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def reverse(l):
+    return l[::-1]
+
+
+def thirds(l):
+    length = int(len(l) / 3)
+    thirds = []
+    thirds.extend(l[length:-length])
```
PythonCHB: same as above, better to use addtion than extend, so it will work with all sequences.

PR #53, opened 2016-10-16
Homework

pvosper:
A quick pass at the rot13 exercise


students/pvosper/session03/rot13.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+print("Caesar Cypher")
+
+alphabet = ("abcdefghijklmnopqrstuvwxyz")
```
PythonCHB: note that there is:
```
In [1]: import string

In [2]: string.ascii_letters
Out[2]: 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ'

In [3]: string.ascii_lowercase
Out[3]: 'abcdefghijklmnopqrstuvwxyz'

In [4]: string.ascii_uppercase
Out[4]: 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
```

students/pvosper/session03/rot13.py
```diff
@@ -0,0 +1,52 @@
...
+
+def reconcile(index):
+    while index not in range(len(alphabet)):
```
PythonCHB: good idea to put docstrings in to say what a function does.

students/pvosper/session03/rot13.py
```diff
@@ -0,0 +1,52 @@
...
+        if index < 0:
+            index += len(alphabet)
+        else:
+            index -= len(alphabet)
+    return index
```
PythonCHB: you could probably use the mod operator this:

In [9]: index % 26
Out[9]: 2

students/pvosper/session03/rot13.py
```diff
@@ -0,0 +1,52 @@
...
+                index = reconcile(alphabet.find(str[i]) + rot)
+                turned = turned + alphabet[index]
+            else:
+                index = reconcile(alphabet.find(str[i].lower()) + rot)
+                turned = turned + alphabet[index].upper()
+        else:
+            turned = turned + str[i]
+    return turned
+
+def test():
```
PythonCHB: great idea to have a test function -- but it's best to put some `assert` statements in there, so you don't have to carefully look at the output to see if it works.

We'll talk about more formal testing soon.

PR #56, opened 2016-10-17
First pass at mailroom homework

darrylsw:
Created data structure and able to add donations and donors to data structure. Able to print thank you note.

Need to work on the following:
- main
  - looping of main menu
- printing thank you note
  - testing for integer value on donation amount
- report
  - sorting of totals
  - formatting printing


students/darrylsw/session02/fizzbuzz.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/local/bin/python3
+
+# created: Darryl Wong
+# date: 10/9/2016
+# Week 2 homework
+
+def fizzbuzz(n):
```
PythonCHB: nice!

students/darrylsw/session02/series.py
```diff
@@ -0,0 +1,96 @@
...
+def lucas(n):
+	"""returns the nth value in the lucas series"""
+	if (n==0):
+		return 2
+	elif (n==1):
+		return 1
+	else:
+		return (lucas (n-1) + lucas (n-2))
+
+def sum_series(n, a=0, b=1):
```
PythonCHB: perfect!

students/darrylsw/session02/series.py
```diff
@@ -0,0 +1,96 @@
...
+assert (sum_series(0, 2, 1) == 2)
+assert (sum_series(1, 2, 1) == 1)
+assert (sum_series(2, 2, 1) == 3)
+assert (sum_series(3, 2, 1) == 4)
+assert (sum_series(4, 2, 1) == 7)
+assert (sum_series(5, 2, 1) == 11)
+assert (sum_series(6, 2, 1) == 18)
+assert (sum_series(7, 2, 1) == 29)
+assert (sum_series(8, 2, 1) == 47)
+
```
PythonCHB: love the asserts!

students/darrylsw/session03/list_lab.py
```diff
@@ -0,0 +1,58 @@
...
+print ("New list")
+print (list)
+
+# prompt the user for a fruit and remove it from the list
+user_input = input ("Enter a fruit to remove from the list: ")
+newlist = []
+for fruit in list:
+	if user_input != fruit:
+		newlist = newlist + [fruit]
+
```
PythonCHB: how would you do it by deleting, rather than putting them all in except that one?

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+
+def main_menu():
+	""" The main menu where everything starts"""
+	print ("Main Menu")
+	print ("1. Send a Thank You")
+	print ("2. Create a Report")
+	print ("3. Exit")
+	return input ("Selection: ")
+
+def print_donors (list):
```
PythonCHB: avoid using "list" -- it is a builtin.

And it's better to use aname that means something -- maybe "donors"?

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+
+def current_donor (list, donor):
+	""" has this donor donated in the past? """
+	for donations in list:
+		if (donor == donations[0]):
+			return True
+	return False
+
+def add_donation (list, donor, amount):
+	""" this donor is an existing donor so we're just adding the amount to the end of his donations """
```
PythonCHB: triple=quoted strings can be multiline -- that's the point :-).

so:
```
    """
    this donor is an existing donor so we're just adding the amount to the end of his donations
    we'll create a new list and just copy elements to new list until we get to our donor and then
    we will modify his donation list
    """
```

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+	""" we'll create a new list and just copy elements to new list until we get to our donor and then """
+	""" we will modify his donation list """
+	new_list = []
+	for donations in list:
```
PythonCHB: no need to do all that copying -- you can find that entry in the list, and then alter it, without changing anything else in the list.

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+		new_list.append (donations)
+	return new_list
+
+def add_donor (list, donor, amount):
+	""" this donor doesn't exist so we'll just add him/her as an element with their donation """
+	#return list + [[donor, amount]]
+	list.append([donor, int(amount)])
+	return list
+
+def print_thank_you_letter (donor, amount):
```
PythonCHB: see if you can use the string format() method to insert the amount and name, rather than building up with +

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+	print (donor + "," + "\n")
+	print ("Thank you for your donation in the amount of $" + amount + ".")
+	print ("Your donation will help our organization in so many ways.")
+	print ("")
+
+def donor_totals (list):
+	""" return an associated array with donation totals for a donor, which is the key for the array """
+	new_list = {}
```
PythonCHB: don't cal lit a list if it is a dictionary (dict) -- what we call associative arrays in Python.

and if you've going to use a dict here, why not for the full data structure?

students/darrylsw/session03/mailroom.py
```diff
@@ -0,0 +1,134 @@
...
+			donation_total = donation_total + donations[i]
+		new_list[donor] = donation_total
+	return new_list
+
+
+if __name__ == '__main__':
+	donor_list = [["Bill Gates", 5000, 10000, 15000], ["Jeff Bezos", 1000, 2000, 3000]]
+	
+	# let's get this puppy started
+	# todo - change to a while loop
```
PythonCHB: good idea :-)

PR #57, opened 2016-10-17
Session 3 Pull Request Including mailroom.py

spencermcghin:
reference mailroom_new.py when providing input please!


codedragon:
I ignored the test.py and the original mailroom.py file. Did comment on the labs. You are off to a good start. Let me know if I can help you figure out how to construct the program as a whole. Seems this might be causing you problems, but I can't tell from the code what exactly you are stumbling on.

PR #60, opened 2016-10-18
session 3 assignments

ninadnaik:
I've attempted to do them (expect mailroom) and mostly got the gist of it. Next step is to cut down the amount of code for the same result.


students/ninadn/class2/class2printcode.py
```diff
@@ -0,0 +1,35 @@
+"""
+Grid printer
+"""
+
+def horzline(grid_size,grid_count,border):
```
PythonCHB: you really do want to use four spaces, and only four spaces, for indentation.

students/ninadn/class3/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+print ('ok\n')
+
+fruitlist = ['mango']+fruitlist
+print (fruitlist)
+
+fruitlist.insert(0,'Banana')
+print (fruitlist)
+print ('ok\n')
+
+#For loop to print only fruits startign with P. I tried using the or in the if statement to check for case, but got an error. Need to investigate.
```
PythonCHB: I usually do something like:

if fruit[0].lower() == 'p'

but you can do:

if fruit[0] == 'P' or fruit[0] == 'p':

or even:

if fruit[0] in 'pP':

students/ninadn/class3/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+
+#Remove the last fruit from the list and display list
+fruitlist.pop()
+print (fruitlist)
+
+#Multiply the list times 2 and get user input for fruit to delete
+fruitlist = fruitlist*2
+userfruit =''
+while userfruit not in fruitlist:
+	userfruit = str(input('Enter a fruit to delete: '))
```
PythonCHB: input always returns a string already

students/ninadn/class3/list_lab.py
```diff
@@ -0,0 +1,84 @@
...
+
+for fruit in fruitlist:
```
PythonCHB: if you delete stuff from the list you are looping through, things can get wonky. better to loop through a copy:

for fruit in fruitlist[:]

students/ninadn/class3/seqfun.py
```diff
@@ -0,0 +1,34 @@
+"""Simple program to take a string and swithc the first and last characters
+
+def switch(word):
```
PythonCHB: you should be able to do all this with slicing!

students/ninadn/class3/seqfun.py
```diff
@@ -0,0 +1,34 @@
...
+	word2 = switch(word)
+	print(word2)
+
+main()
+"""
+
+'''Simple program to return a sequence reversed with slicing only
+
+initstr = str(input("Enter a string :"))
+newstr = initstr[len(initstr):0:-1]+initstr[0]
```
PythonCHB: you rarely need to use len() for this sort of thing.

PR #61, opened 2016-10-18
jhefner: session03

jmhthethird:


students/jhefner/session03/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+"""
+print("\n")
+print("----------------")
+print("Section 4")
+print("----------------")
+items4 = items[:]
+repeat = True
+
+print(items4)
+count=0
```
PythonCHB: tryusing the .count() method of sequences -- strings, lists, tuples all have it.

students/jhefner/session03/list_lab.py
```diff
@@ -0,0 +1,136 @@
...
+Once more, using the list from series 1:
+* Make a copy of the list and reverse the letters in each fruit in the copy.
+* Delete the last item of the original list. Display the original list and the copy.
+"""
+print("\n")
+print("----------------")
+print("Section 5")
+print("----------------")
+items5 = items[:]
+
```
PythonCHB: this all looks good.

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
+#!/Users/jhefner/python_dev/uw_python/bin/python
+import time
+
+verbose = False
+
+
+
+history = {
```
PythonCHB: looks good.

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+"Turanga Leela": [7, 20, 96],
+"Hermes Conrad": [11, 14, 22],
+"Amy Wong": [8, 27, 34],
+}
+
+def send_thank_you():
+    if verbose: print("send_thank_you()")
+    time.sleep(0.5)
+    end = False
+    while not end:
```
PythonCHB: it's a bit cleaner to use `while True:` -- and then `break` when you want to end.

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+        print("--------------------------")
+        print("--------------------------")
+        print("--------------------------")
+        print("Alright lets send that thank you note.")
+        print("Type \"list\" if you need to be reminded of the donors.")
+        print("Otherwise just type the persons full name.")
+        thank_you_name = input("\n>>> ")
+        if thank_you_name == "list":
+            print("Here's the list of our donors:")
+            for i in history: print("> ",i)
```
PythonCHB: do you want to print the whole thing? or just the names?

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if thank_you_name in history:
+            if verbose: print("NAME IS IN LIST")
+        else:
+            history[thank_you_name]=[]
+
+        donation_input = ''
+        print("And what was their donation amount?")
+        donation_input = input("\n>>> ")
+        while not donation_input.isnumeric():
```
PythonCHB: nice find! -- we'll learn a bit better way to do this this week.

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+        end=True
+
+
+
+def create_a_report():
+    if verbose: print("create_a_report()")
+    donation_report = {}
+    for key, value in history.items():
+        value_total = 0
+        for i in value:
```
PythonCHB: this is where the builtin `sum()` function is handy.

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+            value_total+=i
+        donation_report[key] = value_total
+    # print(donation_report)
+    # donation_report = list(donation_report.items())
+    # print(donation_report)
+    # donation_report = sorted(donation_report,key=lambda x:(-x[1],x[0]))
+    # print(donation_report)
+    # ^^^ this is all stuff i found online but i can't find anything that explains how I should be able to manually go over a dictionary or list of tuples.
```
PythonCHB: not sure what you mean by manually -- you can simply loop though a list or tuple (or dict) with "for".

students/jhefner/session03/mailroom.py
```diff
@@ -0,0 +1,99 @@
...
+    donation_report = {}
+    for key, value in history.items():
+        value_total = 0
+        for i in value:
+            value_total+=i
+        donation_report[key] = value_total
+    # print(donation_report)
+    # donation_report = list(donation_report.items())
+    # print(donation_report)
+    # donation_report = sorted(donation_report,key=lambda x:(-x[1],x[0]))
```
PythonCHB: dicts are not ordered -- so you want to put the stuff you need for the report in the list, and then you can sort that list.

PR #63, opened 2016-10-18
Working mailroom

darrylsw:
Functionally mailroom now works, but need to add some more error checking. Bonus functions are there for sorting by number of donations and average donation, but options haven't been added yet when creating the report.


darrylsw:
kata is working, need to string format and get my workaround if it dies prematurely.

PR #66, opened 2016-10-18
Half of session 3 mailroom.py; older assignments 

atruong012:
Uploading older session assignments. Need to revisit Mailroom, as I have the first half finished.


PythonCHB:
I won't comment on the older stuff -- but let me know if you have questions.

students/alextruo/Session3/mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+ #If  user types ‘list’, shows the  list of the donor names and re-prompt
+        if x == "list":
+            #print the list of donors
+            print(donors)
+        elif x in donors:
+#If the user types a name in the list, use it.
+            #tell me how much the current person donated.
+            print("\n",x, "gave this much money:", "$",donors[x])
+            #Ask for another donation amount
+            donate = eval(input("\nDonation amount: " ))
```
PythonCHB: careful! eval() can be dangerous! you pretty much NEVER want to use it.

we'll go over how to test for an int, etc this week.

or, if you really want to evaluate the input, you can use:

import sys

sys.literal_eval(input)

that's a "safe" eval -- it will only evaluate python literals, not arbitrary code.

students/alextruo/Session3/mailroom.py
```diff
@@ -0,0 +1,93 @@
...
+            print_report()
+        elif response == 's':
+            send_thanks()
+        elif response == 'x':
+            break
+        else:
+            print('Please type: "s", "p", or "x"')
+
+if __name__ == "__main__":
+    main()
```
PythonCHB: nice job!

PR #68, opened 2016-10-19
Session 3 Pull Request

heycasey:


students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env python
+
+donors = [('Ed', [1, 4, 5]), ('Thomas', [3]), ('Nancy', [94, 6]), ('Sally', [42, 20, 10, 5, 60])]
+
+
+def print_report():
+    print("This will print a report")
+    for i in range(len(donors)):
```
PythonCHB: this is were you should be able to do:

for donor in donors:
   then_work_with_donor_object

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+        nm = donors[i][0]
+        tot = sum(donors[i][1])
+        avg = find_avg(donors[i][1])
+        count = len(donors[i][1])
+        print("Name is: {name}".format(name=nm))
```
PythonCHB: can you do all these as one format string?

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+        print("Total donation is: {total}".format(total=tot))
+        print("Average donation is: {average}".format(average=avg))
+        print("Total number of donations is: {totalnum}".format(totalnum=count))
+
+
+def send_thanks():
+    print("This will write a thank you note")
+    name_in = input("Enter a name:")
+    for i in range(len(donors)):
```
PythonCHB: again -- can you do: for donor in donors, rather than the index?

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+        if name_in == "list":
+            print(donors[i][0])
```
PythonCHB: after listing, will they get the "Enter a Name:" prompt again? They should.

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+    # Return index number of name
+    for i, donor in enumerate(donors):
+        if donor in donors[i]:
+            return i
+#        return None
+
+
+def find_avg(val):
+    # Return the average of the list
+    return sum(val)/len(val)
```
PythonCHB: may not be worth a function..... good job finding sum() though!

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+            print(donors[i][0])
+        elif name_in in donors[i][0]:
+            don_in = input("Enter a donation amount:")
+            if don_in.isnumeric():
+                donors[name_in][1].extend([int(don_in)])
+
+
+def find_name(name):
+    # Return index number of name
+    for i, donor in enumerate(donors):
```
PythonCHB: yeah enumerate! 

though I think it's better to return the donor object itself, rather than an index...

students/pcasey/session03/mailroom.py
```diff
@@ -0,0 +1,75 @@
...
+            print_report()
+        elif response == 's':
+            send_thanks()
+        elif response == 'x':
+            break
+        else:
+            print('please type "s", "p", or "x"')
+
+if __name__ == "__main__":
+    main()
```
PythonCHB: nice job -- I'm making picky comments -- but it looks like it all works, and working code is good code...

students/pcasey/session03/slice.py
```diff
@@ -0,0 +1,26 @@
+# Slicing lab
+
+a = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
+
+
+def swapends(b):
+    b = a[0], a[-1] = a[-1], a[0]
```
PythonCHB: nice!

though it will only work for mutable sequences -- like lists. Really nice for that, though!

students/pcasey/session03/slice.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def reverse(a):
+    return a[::-1]
+
+
+def midlastfirst(a):
+    b = len(a) // 3
+    c = a[b:-b] + a[-b:] + a[:b]
+    return c
```
PythonCHB: this all looks great!

PR #69, opened 2016-10-19
add mailroom.py again

heb2016:
I have tried numerous times, hopefully this time is successful


students/heb2016/session03/mailroom.py
```diff
@@ -0,0 +1,91 @@
+
+#!/usr/bin/env python
+
+
+def print_report(donorlist):
+    print("This will print a report")
+    row=0
+    while row<=len(donorlist)-1:
```
PythonCHB: you should be able to just loop through the list:

```
for row in donorlist:
    work_with_the_row
   ...
```

students/heb2016/session03/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def input_name():
+        msg ="""
+            Your Name
+            """
+        print(msg)
+        response_name = input("==> ").strip()
+        row =1
+        in_list = False
+        while (row <= len(donorlist)):
```
PythonCHB: easier to use a for loop:

```
for row in donorlist:
    if response_name == row[0]:
        ....
```

better yet, have this in a function, so you can isolate the data structure.

students/heb2016/session03/rot13.py
```diff
@@ -0,0 +1,39 @@
+
+
+def rot13(response):
+        encryp_response=''
+        row=0
```
PythonCHB: why "row" -- it's a good idea to use meaningful names.

students/heb2016/session03/rot13.py
```diff
@@ -0,0 +1,39 @@
...
+        while row <= len(response)-1:
```
PythonCHB: this could be:

```
for char in response:
    ...
```

students/heb2016/session03/rot13.py
```diff
@@ -0,0 +1,39 @@
...
+            if row==0:
+                encryp_respons=response[row]
+            if alpha.rfind(response[row]) ==-1:            
+                encryp_response= encryp_response+response[row]
+                
+
+            elif alpha.rfind(response[row])< (26-13-1):
```
PythonCHB: better to use ord(char) -- it gives you the numerical value of a character.

PR #71, opened 2016-10-20
Homework work

pvosper:


pvosper:
OK, so I found the source of my confusion - I was committing files in the wrong directory. I think I have it sorted now - at any rate, you should see session04 has 5 files so far, including kata_14_trigrams.

pvosper:
I need to include which lab I'm working on in my commit comments :-)

PythonCHB:
hmm -- lots if file changes -- I"ll try to see what's going on.

PythonCHB:
Wow -- this has kinda gone to heck. You have:

IntroPython2016/Session04

rather than it being in your own directory in students.

and a dir called:

students/UWPCE pvosper

and also:

students/pvosper

The goal here is to have everything of yours in:

students/pvosper

you can arrange it inside there however you want, but using teh sessions dirs makes sense.

You should be able to move things around with:

git mv Session04/some_file pvosper/Session04/some_file

Note that git doesn't track directories -- only files. if there is an empty dir, git will ignore it.

Session04/string_formatting_old.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python
```
PythonCHB: yup -- looks like you got much (all) of the class repo's:

Solutions/Session04 

dir in a top-evel dir in yoru repo.

you should just delete these:

```
cd Session04
git rm *
git commit -a
```

and that will help.

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+    ["Ricarda Wykoff", [100, 50, 200]],
+    ["Lue Mollett", [125, 50, 250]],
+    ["Janean Hetzler", [125, 50]],
+    ["Garth Figura", [125]],
+    ["Kurt Cowherd", [100, 50, 25]]
+    ]
+
+# ===== Utilities =====
+
+def donor_name_length():
```
PythonCHB: why do you want to know this?

maybe to know how to print a report?

anyway, the max() function may be useful.

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+    total_donations = 0
+    for entry in donor_list[index][1]:
+        total_donations += entry
+        average_donation = total_donations / len(donor_list[index][1])
+    return average_donation
+
+def donor_total(index):
+    total_donations = 0
+    # BUG: in donor_total - IndexError: list index out of range
+    for entry in donor_list[index][1]:
```
PythonCHB: sum() is handy here

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+def list_donors():
+    for entry in donor_list:
+        print(donor_list[entry][0])
+        
+def find_donor(str):
+    for entry in donor_list:
+        check = entry[0].lower()
+        if check.find(str.lower()) > -1:
+            return donor_list[donor_list.index(entry)]
+    return False # could be: return None
```
PythonCHB: either false or None is fine -- though I'd probably go with None.

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+    return total_donations
+    
+def list_donors():
+    for entry in donor_list:
+        print(donor_list[entry][0])
+        
+def find_donor(str):
+    for entry in donor_list:
+        check = entry[0].lower()
+        if check.find(str.lower()) > -1:
```
PythonCHB: I'm a little confused -- could this just be:

`if check == entry[0].lower():`

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+            return donor_list[donor_list.index(entry)]
+    return False # could be: return None
+
+def second_element(list):
```
PythonCHB: no point to this -- at least not with this name.

IF it were something like:

def donations(donor_record):
    return donor_record[1]

that would let you change the structure of the donor data, and keep the changes isolated.

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+    m   Mail a Thank You letter to a selected donor
+    r   Report list of all donors, sorted by donation size
+    x   Exit the program
+''')
+
+def add_donation():
+    target_donor = []
+    print('''
+    First you'll need to select a donor to add a donation to.
+    Enter a full or partial name to search for:
```
PythonCHB: ahh -- that's what the string.find() was for! nice!

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+    Thank you so much for your generous donation of ${amount:0.2f}.
+    Without such selfless contributions from people like you, we
+    would not be able to continue this important work.
+    We are all very appreciative of your support!
+    
+    Sincerely,
+    
+    Carl
+    '''
+    print(form.format(name=first_name, amount=donations)) # could be: return this
```
PythonCHB: nice use of format!

and yes, I'd probably just return it -- then print it where it's called.

students/UWPCE pvosper/session03/mailroom.py
```diff
@@ -0,0 +1,211 @@
...
+        elif response == "x":
+            print("Exiting program")
+            break
+    
+        else:
+            print("try again")
+
+  
+if __name__ == "__main__":
+    main()
```
PythonCHB: nice job!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+# Read in file
+# temp_file = open('sherlock_partial.txt')
+temp_file = open('sherlock.txt')
+text_string = temp_file.read().replace('\n', ' ')
+temp_file.close()
+
+# Remove upper case
+text_string = text_string.lower()
+
+# Fixups for Punctuation
```
PythonCHB: this is were the string .translate() method is really helpful -- much faster!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+text_string = text_string.replace(' vi ', '')
+text_string = text_string.replace(' vii ', '')
+text_string = text_string.replace(' vii ', '')
+text_string = text_string.replace(' ix ', '')
+text_string = text_string.replace(' x ', '')
+text_string = text_string.replace(' xi ', '')
+text_string = text_string.replace(' xii ', '')
+
+# Fixups for personal names
+text_string = text_string.replace(' holmes ', ' Holmes ')
```
PythonCHB: very nice!

though you might want to have a single list of names tht you can easily add to, then maybe a loop like:

```
for name in names:
    text_string = text_string.replace(name.lower(), name.capitalize())
```

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+# Fixups for days of the week
+text_string = text_string.replace(' monday ', ' Monday ')
+text_string = text_string.replace(' tuesday ', ' Tuesday ')
+text_string = text_string.replace(' wednesday ', ' Wednesday ')
+text_string = text_string.replace(' thursday ', ' Thursday ')
+text_string = text_string.replace(' friday ', ' Friday ')
+text_string = text_string.replace(' saturday ', ' Saturday ')
+text_string = text_string.replace(' sunday ', ' Sunday ')
+
+# Fixups misc contractions and possessives
```
PythonCHB: I think there's a cleaner way to do that. A pain that the apostrophe and the quote are the same character!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+text_string = text_string.replace(' oclock ', " o'clock ")
+text_string = text_string.replace(' mr ', " Mr. ")
+text_string = text_string.replace(' mrs ', " Mrs. ")
+text_string = text_string.replace(' jamess ', " James's. ")
+
+
+# Create list
+text_list = text_string.split()
```
PythonCHB: note that split() splits on any whitespace, so you can keep the '\n's in there if you want.

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+
+# This strips all punctuation but introduces errors
```
PythonCHB: yeah, this is a real pain!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+# for i, entry in enumerate(text_list):
+#     temp_string = ''
+#     for char in entry:
+#         if char.isalpha():
+#             temp_string = temp_string + char.lower()
+#     text_list[i] = temp_string
+
+# $todo another day
+'''
+To find Proper Nouns, look for the NNP tag:
```
PythonCHB: very cool! does is look like that'll work?

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+
+from nltk.tag import pos_tag
+
+sentence = "Michael Jackson likes to eat at McDonalds"
+tagged_sent = pos_tag(sentence.split())
+# [('Michael', 'NNP'), ('Jackson', 'NNP'), ('likes', 'VBZ'), ('to', 'TO'), ('eat', 'VB'), ('at', 'IN'), ('McDonalds', 'NNP')]
+
+propernouns = [word for word,pos in tagged_sent if pos == 'NNP']
+# ['Michael','Jackson', 'McDonalds']
+'''
```
PythonCHB: if you've identified the proper nouns, maybe you could do something with them -- like note in the trigram that there is a proper noun in the word pair, and substitute a random name for the actual name..

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+
+# In trigrams dictionary, each value is a list of lists
+trigrams = {}
+
+# Build dictionary of trigrams
+for i in range(len(text_list)-2):
+    # I could make this one line, but then I would understand it even less
```
PythonCHB: good call!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+    key = text_list[i] + " " + text_list[i+1]
+    val = [text_list[i+2]]
+    trigrams.setdefault(key, []).append(val)
+
+def pick_value(val_list):
```
PythonCHB: there is a random.choice() method that randomly picks something out of a list.

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+    # random includes upper range value
+    return val_list[random.randint(0, len(val_list) - 1)]
+    
+def make_sentance():
+    sentance = ""
+    seed = random.choice(list(trigrams.keys()))
```
PythonCHB: ah -- you found it!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+    sentance_list = seed.split()
+    for i in range(random.randint(8, 24)):
+        seed = sentance_list[-2] + " " + sentance_list[-1]
+        sentance_list.append(pick_value(trigrams[seed])[0])
+    for word in sentance_list:
+        sentance = sentance + " " + word
+    sentance = sentance[1].upper() + sentance[2:] + '.'
+    return sentance
+    # $todo don't end sentences with joins like 'and', 'the', 'a'
```
PythonCHB: now you are really getting fancy!

students/UWPCE pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
...
+    for i in range(random.randint(3, 7)):
+        paragraph = paragraph + " " + make_sentance()
+    return paragraph
+
+# 3 to 5 sentences per paragraph
+
+if __name__ == '__main__':
+    print('\n=== MAIN ===\n')
+    
+    print(make_paragraph())
```
PythonCHB: this is really nice work -- looks like you enjoyed really thinking about it.

students/pvosper/session04/kata_14_trigrams.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env python3
```
PythonCHB: this seems to be another copy...

students/UWPCE pvosper/session03/mailroom.py
pvosper: I used that to determine the maximum width of the name for formatting the report, rather than a fixed number.

students/UWPCE pvosper/session03/mailroom.py
pvosper: I hit that IndexError once, but never saw it again. It could be that something was wonky in iPython?

PR #74, opened 2016-10-22
dict_lab v1

heycasey:
Can't figure out why the "for" loop will only count the number of "t"s in just the first dict entry.


PythonCHB:
last for loop in the dict lab? or???

PythonCHB:
I don't see mailroom -- did you remember to "git add" it?

PR #77, opened 2016-10-25
Catching up with some stuff

komashu:
Not finished with all of it.


students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
...
+class TestOfMathFunctions:
+
+    def test_fibonacci(self):
+        pass
+
+    def test_lucas(self):
+        pass
+
+    def test_sum_series(self):
+        pass
```
codedragon: Testing is a great idea. We'll start testing very soon. :)

students/shulatif/session_02/series.py
```diff
@@ -0,0 +1,77 @@
...
+                randos = n
+            elif count == 1:
+                randos = x
+            else:
+                randos = n + x
+                n = x
+                x = randos
+            print(randos)
+            count += 1
+
```
codedragon: The instructions may not have been clear enough. The idea was to notice how similar the fibonacci code and the lucas code were, and then construct one function that would return either the value from fibonacci or lucas, depending on the inputs. It would not actually call the already created code, but it would be new code that could return one or the other type series.

codedragon:
Off to a good start. Mailroom is very nice.

students/shulatif/session_02/series.py
komashu: oh. i didn't realize that. was trying to use the DRY principle.

PR #78, opened 2016-10-25
A smattering of homework. 

bc-uw:


students/bc-uw/session03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+        if response not in donors.keys():
+            donors[response] = []
+        else:
+            try:
+                newdonation = int(input("Enter Donation Amount: "))
+            except ValueError:
+                print("Please enter an integer")
+                pass
+            donors[response].append(newdonation)
+
```
codedragon: Here is a great place for setdefault. Instead of checking to see if response is already in donors.keys, you can use ```donors.setdefault(response, []).append(newdonation)```  
This will check if response is in donors, and if it is not, will create the key/value pair with the empty list as value, and then appends the newdonation to the list. If it does exist, it ignores setdefault, and just adds to the existing lsit.

students/bc-uw/session03/mailroom.py
```diff
@@ -0,0 +1,74 @@
...
+    #get values of list items in dict, add to list
+    for k in donors.keys():
+        for v in donors[k]:
+            mylist.append(v)
+
+    total_donations = sum()
+    number_donations = len()
+    average = total_donations / number_donations
+    report.append(donor, total_donations, number_donations, average)
+
```
codedragon: You are on the right track here. You need to go through the donors, and for each one, look at the donations, get the total, the number, average. So your code for that should be inside the loop. To get both the keys and the values from the dictionary, use items()
```
for donor, value in donors.items():
    total_donations = sum(value)
    number_donations = len(value)
    ave_donations = total_donations / number_donations
    report.append(donor, total_donations, number_donations, ave_donations)
```

students/bc-uw/session03/slice.py
```diff
@@ -0,0 +1,15 @@
+
+def reverse_ends(string):
+    return string[-1] + string[1:-2] + string[0]
+
```
codedragon: Remember this only works for strings!

students/bc-uw/session04/file-lab.py
```diff
@@ -0,0 +1,18 @@
...
+            dest.write(line)
+
+with open("students.txt") as students:
+          langlist = []
+          for line in students:
+                  langs = line.rsplit(':')[1]
+                  langlist.append(langs)
+          strippedlist = [i.strip() for i in langlist]
+          noquotelist = [i.replace("'", "").replace(",", "") for i in strippedlist]
+          uniqueset = set(noquotelist)
```
codedragon: Nice list comprehensions!

codedragon:
I had to make a change to the gitignore, which I hope will not cause problems...

PR #79, opened 2016-10-25
Session 4 and Session 3 homework

atruong012:
String formatting, list lab, dictionary lab, file lab (minus send part), updated mailroom file with dictionary that finally works. :)


students/alextruo/Session3/list_lab.py
```diff
@@ -0,0 +1,90 @@
...
+    #ask if the user likes the fruit in the list
+        favorite = input("Do you like: {name}? yes or no: ".format(name=item))
+        if favorite == 'no':
+            while item in fruit:
+                fruit.remove(item)
+                print("We are removing it.")
+        elif favorite == 'yes':
+            #skip to the next item in the list
+            pass
+        else:
```
codedragon: Since you aren't doing anything with the favorite == 'yes' clause, you can just skip it and use ```elif favorite != 'yes' instead of your else statement.

codedragon:
You are getting the hang of things. Pay attention to when/why a while loop is necessary, and try replacing with if statements.

PR #80, opened 2016-10-25
adding session 3

Baumelbi:


students/baumel/session_03/list_lab.py
```diff
@@ -0,0 +1,65 @@
...
+new_fruit1 = input("Enter another Fruit: ")
+fruit = [new_fruit1] + fruit
+print(fruit)
+#Add another fruit to the beginning of the list using insert() and display the list.
+new_fruit2 = input("Enter another Fruit: ")
+fruit.insert(0,new_fruit2)
+print(fruit)
+#Display all the fruits that begin with “P”, using a for loop.
+length = len(fruit)
+for i in range (0,length):
```
PythonCHB: you hardly ever need to lopo through indexes:

```
for f in fruit:
    if f[0].lower() == 'p':
        print(f)
```

students/baumel/session_03/list_lab.py
```diff
@@ -0,0 +1,65 @@
...
+#Display the list.
+print(fruit)
+#Ask the user for a fruit to delete and find it and delete it.
+fruit_remove = input("Enter a Fruit to remove: ")
+fruit.remove(fruit_remove)
+print(fruit)
+
+"""Part 3 of Lab"""
+#Ask the user for input displaying a line like “Do you like apples?”
+for i in fruit:
```
PythonCHB: not great to call it "i" -- but this is the right idea.

PythonCHB:
How are you doing the the newer assignments?

PR #81, opened 2016-10-26
Class work and Labs

simbachange:


simbachange:
Old commits that i had not pulled.. sorry

PR #83, opened 2016-10-26
Session04 Labs and Assignments

psbriant:
Please look at line 93 in kata_fourteen.py. Getting an error between lines 80 and 93. Please ignore modifications to mailroom.py in session03.


students/psbriant/session04/kata_fourteen.py
```diff
@@ -0,0 +1,129 @@
...
+    Once entire file has been iterated through:
+    Choose random word pair as the start and use these to look up the next
+    word.
+    """
+    rand_item = random.choice(str_list)
+    item2_index = str_list.index(rand_item) + 1
+    item2 = str_list[item2_index]
+    start_key = rand_item + ' ' + item2
+    return start_key
+
```
codedragon: What happens if you get the last word in your list as your random choice? One way to avoid this, while reducing the line count and hopefully even making things clearer:
```
index = random.choice(range(0, len(str_list) - 1))
return str_list[index] + ' ' + str_list[index + 1]

File: students/psbriant/session04/kata_fourteen.py [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+    Use it to lookup the next string.
+    Split the first string and use the second and third strings to look up the
+    next string.
+    If a key has multiple values, choose a random value.
+    """
+    # Strings to write to file
+    out_list = []
+    # Current key
+    key_phrase = str_key
+
```
codedragon: Why do you change the name of str_key? Why not just name it what you want in the first place?

students/psbriant/session04/kata_fourteen.py
```diff
@@ -0,0 +1,129 @@
...
+        if key_phrase not in tri_dict:
+            break
+        # Add element to out list
+        out_list = out_list.append(key_phrase)
+        # Get the next string
+        next_str = tri_dict[key_phrase]
+        if len(next_str) > 1:
+            # Add random element to list
+            out_list = out_list.append(random.choice(next_str))
+        else:
```
codedragon: As we discussed, you should just append, not set a new variable. Append is an in place operation, and will not return anything. It takes a while to figure out which ones are in place and not, but you can always test in the interpreter. In place operations return None:
```
In [41]: my_list
Out[41]: ['I', 'am', 'struggling,', 'but', 'will', 'succeed', 'whatever']

In [42]: test = my_list.append('ack')

In [43]: print(test)
None
In [44]: my_list
Out[44]: ['I', 'am', 'struggling,', 'but', 'will', 'succeed', 'whatever', 'ack']
```
Notice test is set to None, but my_list now has ack in it.

students/psbriant/session04/kata_fourteen.py
```diff
@@ -0,0 +1,129 @@
...
+            # Add to list
+            out_list = out_list.append(next_str)
+        # Assign new strings to key phrase
+        key_phrase = key_phrase.split
+        key_phrase = key_phrase[1] + ' ' + next_str
```
codedragon: You already know what is wrong here...

students/psbriant/session04/kata_fourteen.py
```diff
@@ -0,0 +1,129 @@
...
+    string and write to a new file.
+    """
+    text = " ".join(trigram)
+    with open('new_trigram.txt', 'w') as outfile:
+        outfile.write(text)
+    outfile.close()
+
+# ==============================================================================
+
+
```
codedragon: This would of course be more useful if you could input the file names when you call your script, rather than hard-coding it, but probably you won't be using this script much after this. ;-) It is something that will come up again in any case.

PR #84, opened 2016-10-26
Session 4 labs

charlesrobison:
Unable to finish


students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+for item in os.listdir(cwd):
+    print(cwd + "/" + item)
+
+# write a program which copies a file from a source, to a
+# destination (without using shutil, or the OS copy command)
+file = open('file_lab01.txt', 'r')
+file_text = file.read()
+file_new = open('file_lab02.txt', 'w')
+file_new.write(file_text)
+file.close()
```
PythonCHB: better to use "with" than have to close the file by hand.

students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+file_new.close()
+
+# advanced: make it work for any size file: i.e. don’t read
+# the entire contents of the file into memory at once.
+with open('file_lab01.txt','r') as r, open('file_lab02.txt', 'w') as w:
```
PythonCHB: perfect!

students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+    for line in r:
```
PythonCHB: this will only work for txt files -- but it is a fine solution in that case.

students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+        w.write(line)
+r.close()
```
PythonCHB: no need to call close() when you have used "with" -- at the end of the with block, the files are closed -- that is (most of) the point of with.

students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Write a little script that reads that file, and generates a list
+# of all the languages that have been used.
+languages = []
+with open('students_test.txt', 'r') as infile:
+    for line in infile:
+        if not ':' in line:
+            continue
+        for student, language in line.split(':'):
+            languages.append(language)
```
PythonCHB: there may be more than one language after the colon -- so you need to parse that our too.

students/crobison/session04/file_lab.py
```diff
@@ -0,0 +1,47 @@
...
+print(languages)
+# Getting value error:
+# ValueError: too many values to unpack (expected 2)
```
PythonCHB: that would happen if there were more than one colon on a line -- which I don't think there are, but maybe ??

to debug, you want to print each line you are processing, so you can see what line it fails on.

keep in mind that split() will return from one to any number of results. but you can pass an extra argument saying you want it to only spilt on the first one..

In [12]: s = "this: that: the other"

In [13]: s.split(":")
Out[13]: ['this', ' that', ' the other']

In [14]: s.split(":",maxsplit=1)
Out[14]: ['this', ' that: the other']

PR #87, opened 2016-10-30
Homework Review: File Lab, Exceptions Lab & Comprehensions Lab

pvosper:
Finished labs.


students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
-print(os.getcwd(), ":")
+print('\nFull path to current directory:')
+print(os.getcwd())
 
+# create list of all files in current directory
 directory_list = os.listdir()
 
+print('\nFiles in current directory:')
 for entry in directory_list:
     print("\t", entry)
```
PythonCHB: ooo -- fancy!

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+def line_of_gibberish():
+    import random
+    line = ""
+    for i in range(80):
+        line = line + chr(random.randint(41,79)) #$Todo: hex
+    line = line + "\n"
+    return line
+
+# Create file to use as source
+file_object = open('source_file.txt', 'w')
```
PythonCHB: nice -- you are your way to formal testing!

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+for i in range(100):
+    file_object.write(line_of_gibberish())
+file_object.close()
+
+# Copy source to target line by line
```
PythonCHB: This is good -- but only for text files -- how would you do it for binary files?

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+target_file_object = open('target_file.txt','w')
+source_file_object = open('source_file.txt','r')
+while True:
+    line = source_file_object.readline()
+    if not line:
+        break
+    else:
+        target_file_object.write(line)
+target_file_object.close()
```
PythonCHB: better to use "with" -- so you don't need to worry about closing the files.

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+# Create a dictionary from students.txt file
+# key = name, value = languages
+students_dictionary = {}
 
 for line in open('../../../Examples/Session01/students.txt'):
-    s = []
-    s = line.split(':')
-    if len(s) == 2:
-        d[s[0]] = s[1].replace('\n', ' ')
+    student_info = []
```
PythonCHB: non need to initialize an empty list -- split() returns a new list anyway.

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
-
-# .strip().lower()
-
-# f = open('myfile','w')
-# f.write('hi there\n') # python will convert \n to os.linesep
-# f.close() # you can omit in most cases as the destructor will call it
+        students_dictionary[student_info[0]] = '<n/a>'
+
+# Get rid of 'name: languages' header line
+students_dictionary.pop('name')
```
PythonCHB: This works fine-- but why not just not read that line of the file in the first place?

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+
+# Create a dictionary of languages
+# key = language, value = count
+# $todo value = list of students who know that language
+language_dictionary = {}
+language_list = students_dictionary.values()
+for entry in language_list:
+    entry = entry.lower().replace(',', '').replace('pyton', 'python').replace('macscript', 'maxscript')
```
PythonCHB: nice!

students/pvosper/session04/file_lab.py
```diff
@@ -1,59 +1,157 @@
...
+	 file_lab.py
+	 kata_14_trigrams.py
+	 lorem_ipsum_generator.py
+	 myfile
+	 sherlock.txt
+	 sherlock_partial.txt
+	 source_file.txt
+	 target_file.txt
+	 temp_file.txt
+
```
PythonCHB: this all looks good.

students/pvosper/session05/comprehensions_lab.py
```diff
@@ -0,0 +1,291 @@
...
+count_evens([2, 2, 0]) == 3
+
+count_evens([1, 3, 5]) == 0
+
+def count_evens(nums):
+   one_line_comprehension_here
+'''
+
+def count_evens(list_of_numbers):
+    result = len([evens for evens in list_of_numbers if evens % 2 == 0])
```
PythonCHB: think about naming here -- "evens" in this context will be all the numbers before it hits the "if" filter.  Also, "evens" is plural, but is being used for a single number. Inside a  list comp, not a big deal, but better to use a name like "num" (or even just "n").

students/pvosper/session05/comprehensions_lab.py
```diff
@@ -0,0 +1,291 @@
...
+
+print('\n', food_prefs_string.format(
+    food_prefs['name'], 
+    food_prefs['city'],
+    food_prefs['cake'],
+    food_prefs['fruit'],
+    food_prefs['salad'],
+    food_prefs['pasta'],
+    ))
+
```
PythonCHB: this is a good chance to use "named" format specifiers:

"{name} is from {city},......"

then you can pas the whole dict in like so:

.format(**food_prefs)

we'll get into that whole ** thing soon!

students/pvosper/session05/comprehensions_lab.py
```diff
@@ -0,0 +1,291 @@
...
+2. Using a list comprehension, build a dictionary of numbers from zero to 
+fifteen and the hexadecimal equivalent (string is fine). (the hex() function 
+gives you the hexidecimal representation of a number.)
+
+Do the previous entirely with a dict comprehension – should be a one-liner
+
+'''
+
+hex_dictionary = { i: hex(i) for i in range(16) }
```
PythonCHB: bingo!

students/pvosper/session05/comprehensions_lab.py
```diff
@@ -0,0 +1,291 @@
...
+for j in range(2,10):
+    dictionary_of_sets[j] = {i for i in range(21) if i % j == 0}
+print('\ndictionary_of_sets:\n', dictionary_of_sets)
+
+# Create a list of divisible sets as a single-line comprehension
+list_of_sets = [{i for i in range(21)if i % j == 0} for j in range(2,10)]
+print('\nlist_of_sets\n', list_of_sets)
+
+# Create a dictionary of divisible sets as a single-line comprehension
+dictionary_of_sets = {j: {i for i in range(21) if i % j == 0} for j in range(2,10)}
```
PythonCHB: nice!

students/pvosper/session05/exceptions_lab.py
```diff
@@ -0,0 +1,106 @@
...
+->^C---------------------------------------------------------------------------
+KeyboardInterrupt                         Traceback (most recent call last)
+<ipython-input-3-f9c05402e592> in <module>()
+----> 1 response = input('->')
+
+KeyboardInterrupt: 
+
+'''
+
+def safe_input():
```
PythonCHB: in real life, you wouln'dt have the prints -- but yes, that's the idea.

Also -- if you are going to do the same thing with two Exceptions, you can do:

except  (EOFError, KeyboardInterrupt):
    ....

PR #89, opened 2016-10-30
Session04+: Please take a peek at trigram if you could

jmhthethird:
Heya ,

If you could take a look at my trigram.py and let me know what you think of the solution that would be great. It was a good exercise for sure. 

Thanks!


students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
 #!/Users/jhefner/python_dev/uw_python/bin/python
 
-import os
+import json #just used to pretty print the dictionary when I was debugging
```
PythonCHB: interesting choice!

did you try the pprint module?

```
from pprint import pprint

pprint(something)
```
gives you a prettier print out of any python object.

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
 """
 Thoughts on how im going to do this:
+---
 How to store the data: a dictionary.
 Key = First two words (starting at position 0,1 in line of document and incrementing to end of line
-Value = set. set is populated with each the 2 words that come after the key. This will retain only unique matches without incurring overhead of duplicate checking.
+Value = list. list is populated with each the 2 words that come after the key. This will retain only unique matches without incurring overhead of duplicate checking.
 
 The method for creating the dictionary:
 Bring entire file into memory if possible. If it looks like the file is too big I'll have to evaluate it line by line.
```
PythonCHB: I wouldn't worry about that at this point.

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
 2. Take Key and append to output.
-3. Use rand to choose a random value from keys set and append to output.
+3. Use rand to choose a random value from keys list and append to output.
 4. Evaluate last word in output, find rand dictionary key sthat starts with the word.
 5. Step 2-4 until done.
-
-
 """
 
 def read_file_to_memory(filename):
```
PythonCHB: there isn't much point in making this a function -- it's a two liner:

```
with open(filename) as f:
          file_data = f.read()
```

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
-    print(dict_of_words)
-
-
+def create_trigram_dictionary(file_data):
+    listerizer = file_data.split()
+    trigram_dictionary = {}
+    for i in range(len(listerizer)-2):
+        key = listerizer[i] + " " + listerizer[i+1]
+        val = listerizer[i+2]
+        trigram_dictionary.setdefault(key,[]).append(val)
```
PythonCHB: perfect!

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
+    return trigram_dictionary
 
+def create_trigram_book(trigram_dictionary):
+    random_starting_key = random.choice(list(trigram_dictionary.keys()))
```
PythonCHB: too bad that you need to make the whole list to take a sample, but you do....

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
+    words_list = " ".join((random_starting_key,random.choice(trigram_dictionary[random_starting_key]))).split()
+    phrase = " ".join((words_list[-2],words_list[-1]))
+    # print("word list:",words_list)
+    # print("phrase:",phrase)
+    # print("phrase type:",type(phrase))
+    # print("phrase in trigram dict:",phrase in trigram_dictionary)
 
+    while phrase in trigram_dictionary:
```
PythonCHB: isn't this going to go on potentially forever?

students/jhefner/session04/trigram.py
```diff
@@ -1,53 +1,69 @@
...
+        words_list.append(random.choice(trigram_dictionary[phrase]))
```
PythonCHB: I'm seeing repeated code here --- not a lot, but DRY! (Don't Repeat Yourself).

why not start your words_list with a random pair from the dict, then go into the loop, isng the last two words in the list each time through the loop.

students/jhefner/session05/list_comprehension.py
```diff
@@ -0,0 +1,113 @@
...
+    # ok im not quite sure whats gonna happen here. if i had to guess
+    # its going to turn each element of each list of items into a tuple of the 1st and 2nd element from that list.
+    # so comprehension would become a combination of both elements of eggs/meats and look sorta like:
+    #  comprehension = ('poached egg', 'fried egg'), ('lite spam', 'ham spam').....fuu actually im not sure...
+    # this one has 3 elements...
+
+    print(comprehension[0])
+    # no idea...
+
+    # OHHHh so it was doing basically what ZIP was doing the other day.
```
PythonCHB: not quite -- try zip with these two lists -- you'll see a difference!

PR #91, opened 2016-10-31
master

spencermcghin:


students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
-import numpy
+import numpy, collections
 
+from tabulate import tabulate
```
PythonCHB: nice find -- I'd never seen that.

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -1,7 +1,8 @@
...
 
 """Dictionary of Donors and Amounts Donated"""
```
PythonCHB: better to use comments for things like that:

```
# Dictionary of Donors and Amounts Donated
```
keep the triple quoted strings for docstrings - right after function definitions, etc.

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -59,15 +57,15 @@ def add_donor():
 # Add donation amount to new donor #
 
 def add_amount():
-    donors.update({# Variable : donation_amount})
+    donors.update({# Variable) : int(donation_amount)})
```
PythonCHB: this is currently a syntax error:

```
$ python3 mailroom_new.py 
  File "mailroom_new.py", line 64
    def check_donation():
      ^
SyntaxError: invalid syntax
```

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -59,15 +57,15 @@ def add_donor():
...
 
 # Verify donation amount is an integer #
 
 def check_donation():
     while isinstance(donation_amount, int):
```
PythonCHB: it will never be an instance of int unless it successfully converted to int already.

now that you now Exceptions, better to do something like:

```
try:
    amount = float(input_amount):
except ValueError:
    print("not valid -- try again")
```

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -87,35 +85,29 @@ def print_email():
...
-        totals_results.append((sum(donations)))
-
-# Print total number of donations #
-
-def print_num_donations():
-    num_results = []
+def report_data():
+    # establish separate dictionary objects #
+    results = collections.OrderedDict()
+    donor_dict = {"Donors": []}
```
PythonCHB: looking good.

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -87,35 +85,29 @@ def print_email():
...
+        num_results["Number of Donations"].append(len(donations))
+        avg_results["Average Donation"].append(int(numpy.mean(donations)))
+    # combine dictionary objects into one for tabulate data input format #
+    results.update(donor_dict)
+    results.update(totals_dict)
+    results.update(num_results)
+    results.update(avg_results)
+    return results
+
+print(tabulate(report_data(), headers="keys", tablefmt="fancy_grid", numalign="center"))
```
PythonCHB: did this end up working out well?

students/spencer_mcghin/session4/Files_Lab.py
```diff
@@ -0,0 +1,26 @@
...
+def print_path():
+    for file in os.listdir(os.getcwd()):
+        print(os.path.abspath(file))
+
+# Read students file and generate languages list #
+
+def gen_lang():
+    doc = list(open('/Users/SMcGhin/Documents/IntroPython2016/Examples/Session01/students.txt', 'r'))
+    data = []
+    for line in doc[1:-1]:
```
PythonCHB: if you are going to loop thorugh the lines in a text file, easiest to do:

```
with open( .....) as doc:
    for line in doc:
        ....
```

If you need to skip the header, you can do a .readline() call before the the loop.

students/spencer_mcghin/session4/Files_Lab.py
```diff
@@ -0,0 +1,26 @@
...
+        students, languages = line.strip().split(':')
+        data.append(languages)
```
PythonCHB: you'll get a lot of repetition of languages -- how would you deal with that?

students/spencer_mcghin/session4/sherlock_trigram.py
```diff
@@ -0,0 +1,8 @@
+def trigram():
+    text = open('/Users/SMcGhin/Documents/IntroPython2016/students/spencer_mcghin/session4/sherlock_small.txt', 'r')
+    for words in text:
```
PythonCHB: when you loop this way, you get the file line by line.

so no point in splitting on a newline.

PR #92, opened 2016-11-01
working trigram

darrylsw:
working trigram homework
assorted lab assignments.


codedragon:
For your mailman project, it looks like you have chosen a list of lists for your data structure. 
What are the other structures you could have chosen? Would any of them simplified your code, especially to get rid of so many for loops through the data.

students/darrylsw/session04/kata.py
```diff
@@ -0,0 +1,92 @@
...
+				#print ("key: {} value: {}".format (key, line[0]))
+			else:
+				#print (line[i-1], line[i], line[i+1])
+				key = (line[i-1], line[i])
+				#print (key)
+				dict[key] = line[i+1]
+				#print ("key: {} value: {}".format (key, line[i+1]))
+		next2last = line[len(line)-2]
+		last = line[len(line)-1]
+	f.close()
```
codedragon: I would try to change this giant loop into smaller pieces. First pull the data from the file, and then parse. Your data is not actually dependent on the lines, and this seems to be making your parsing more complicated, so consider reading the file like this:
```
>>> with open('workfile', 'r') as f:
...     read_data = f.read().split()
```
This will close the file automatically after reading the data and putting the data into one list (gets rid of formatting like '\n'). Now you have the data, and can parse through to get your words.

students/darrylsw/session04/kata.py
```diff
@@ -0,0 +1,92 @@
...
+				#print (next2last, last, line[0])
+				key = (next2last, last)
+				dict[key] = line[0]
+				next2last = ""
+				#print ("key: {} value: {}".format (key, line[0]))
+			else:
+				#print (line[i-1], line[i], line[i+1])
+				key = (line[i-1], line[i])
+				#print (key)
+				dict[key] = line[i+1]
```
codedragon: What happens when you find a word pair a second time? Also, please do not use dict as a variable name, since that is a key word in python.

students/darrylsw/session04/kata.py
```diff
@@ -0,0 +1,92 @@
...
+		first = initial_key[0]
+		second = initial_key[1]
+		third = dict[initial_key]
+	
+		print ("{} {} {}".format(first, second, third), end=" ")
+		return (second, third)
+	else:
+		key = (first, second)
+		value = dict[key]
+		print ("value: ", value)
```
codedragon: Something must not be quite right with your dictionary or your method of calling it. You should not have keys with no value. When you are creating the dictionary, you should not be adding a key unless there is a word that follows the two words in the key in the original text. And since the dictionary values of any particular pair of words are words that follow that pair of words in the text, there will always be at least one other key that includes the last word of your previous key and one from the list of values from that key. The randomness comes from having a list of words that follow a particular pair (you can imagine 'and the' having many words that follow it in various places), and randomly choosing a word from that list for your new text.

students/darrylsw/session05/exceptions_lab.py
```diff
@@ -0,0 +1,32 @@
...
+		an_input = input (string)
+	except EOFError:
+		print ("received EOFError")
+		return None
+	except KeyboardInterrupt:
+		print ("received KeyboardInterrupt")
+		return None
+	else:
+		return an_input
+
```
codedragon: This is good! Did you know you could also do:
```
try:
    return input(string)
except
...
```
and eliminate the else? Good practice learning what the else does in this case, though.

PR #93, opened 2016-11-02
triagram_lab.py from Session04 and mailroom2.py from Session 05

nachiketgalande:
Finished the triagram lab from last week and added "safe_input" method to mailroom2.py


codedragon:
I made a bunch of comments, but this is very good work. :)

PR #94, opened 2016-11-02
Session 05

psbriant:
Exceptions Lab and Comprehensions Lab. Please take a look at everything except Trigrams. Maria has already provided feedback for Trigrams.


codedragon:
More good work Paul.

PR #95, opened 2016-11-03
Week 6 Assignments

jerbear4:


students/jbearer/session05/exception_lab.py
```diff
@@ -9,7 +9,8 @@ def wrapper():
     try:
         userInput = input('Enter ^C or ^D: ')
     except (KeyboardInterrupt, EOFError) as the_error:
-        print('\nNone')
+        return None
+    return userInput
 
-wrapper()
+print(wrapper())
```
codedragon: Oops, should have caught this earlier. If you are going to catch an exception, and not re-raise it, you should give feedback to your user (and usually log the exception). In this case, it is not so important, of course, but good habit to get into.

students/jbearer/session06/mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+
+def proc_donation(d_name, donation):
+    if d_name in donor_list:
+        donor_list[d_name].append(donation)
+    elif d_name not in donor_list:
+        donor_list[d_name] = [donation]
+    print()
+    print('Thank you {} for your donation of ${}!!!'.format(d_name,donation))
+    print()
+
```
codedragon: You have quite a few functions that have printing in them. Try to separate the logic from the formatting. Iow, have functions that just doing logic, and functions that are just doing output. So, for example, in proc_donations, you could have it return the new donor_list and then it can call another function that does the actual printing. That way you can have a test that just makes sure that the donation is added to the list, and that your print function is called with the correct donation, without having to capture output to make sure something was actually printed, which is kind of a pain.

students/jbearer/session06/test_mailroom.py
```diff
@@ -0,0 +1,16 @@
...
+from mailroom import print_report, proc_donation, get_donation
+
+def test_print_report_1():
+    assert print_report() is None
+
+def test_proc_donation_1():
+    assert proc_donation('Fred', 269) is None
+
+def test_get_donation_1():
+    get_donation()
```
codedragon: Off to a good start on testing. The key to testing is to look at each function, and figure out what it is suppose to be doing. Then write a test that makes sure it is doing that thing. Sometimes you will discover that a function is doing multiple things, and often this means you should break it into a couple of functions, making it easier to read and test. Is the function of print_report really to return None? Print_report is really doing several things, so maybe it should be broken apart. The key things it is doing is creating sort_l, so maybe you should have a function that is just creating sort_l, and then you can run a test on it to make sure it has all of the donors you expect, starts with the correct donors (sorted correctly, includes columns expected, etc.) Then send this variable to a function that literally just prints stuff. Generally, we are much less concerned about testing formatting with unittests, as this is something that often just needs to be done visually. Let me know if this raises more questions than it answers...

PR #96, opened 2016-11-06
For review: test_lab; WIP: key_lab, mailroom_2

pvosper:
**key_lab** and **mailroom_2** should be ready for non-class Tuesday; plus working on **Lightning Talk** for 15th


students/pvosper/session06/test_lab.py
```diff
@@ -0,0 +1,58 @@
...
+    assert count_evens([2, 1, 2, 3, 4]) == 3
+
+def test_count_evens_b():
+    assert count_evens([2, 2, 0]) == 3
+
+def test_count_evens_c():
+    assert count_evens([1, 3, 5]) == 0
+
+# close_far   
+
```
PythonCHB: this all looks great.

now that you have tests for close_far, maybe see if you can refactor it -- that looks like a lot of code. (not that I have anything in mind, but I'll bet it could be made cleaner).

And it's a great way to appreciate the benefits of unit testing.

PR #98, opened 2016-11-07
Week 6.5 Assignments

jerbear4:
Reworked mailroom.py to separate print and logic functions.  Wrote pytest against logic functions of mailroom.py.  Not sure if I am doing it correctly.  The functions do not return anything so the return is 'None'.  Unless I am doing something wrong.  Please review.

### DIFF
--- a/students/jbearer/session06/test_mailroom.py
+++ b/students/jbearer/session06/test_mailroom.py
@@ -4,13 +4,10 @@ Excercise: Write unit tests for mailroom.py
 November 2, 2016
 """
 
-from mailroom import print_report, proc_donation, get_donation
+from mailroom import sort_donors, proc_donation
 
-def test_print_report_1():
-    assert print_report() is None
+def test_sort_donors():
+    assert sort_donors() is None
 
 def test_proc_donation_1():
-    assert proc_donation('Fred', 269) is None
-
-def test_get_donation_1():
-    get_donation()
+    assert proc_donation('Fred', 35) is None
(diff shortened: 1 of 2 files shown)


PythonCHB:
yup -- test functions don't return anything -- if an Error is raised or an Assertion fails, then test fails -- otherwise it passes.

students/jbearer/session06/test_mailroom.py
```diff
@@ -4,13 +4,10 @@
 November 2, 2016
 """
 
-from mailroom import print_report, proc_donation, get_donation
+from mailroom import sort_donors, proc_donation
 
-def test_print_report_1():
-    assert print_report() is None
+def test_sort_donors():
+    assert sort_donors() is None
```
PythonCHB: I see what you mean now -- 

If you have a function that doesn't return anything, it is indeed hard to test. So this test is notn completley useless, it does test that the function runs without error -- a good start.

But a test really should test that that logic of the function was performed correctly. So you need to see if the donors really are sorted after this function is called.

As I mentioned in class, if it's hard to test, it may mean it needs refactoring:

Could sort_donors() return the now sorted donor list?

or, if a function changes state of some object, examine that object to see if the state was changed properly. Presumably SOMETHING is supposed to be sorted after this function is called -- the the test choudl check that it is.

as for print_report() -- hard to test printing (not impossible, but we don't need to go there yet :-) ) -- but what you could di is have a function that generates the report as a string and returns that -- you can then test that. the actually printing doesn't need testing -- we can assume Python's print function works :-)

PR #99, opened 2016-11-08
Home work from Session5 & Session6

heb2016:
Commit four files:

git commit -m 'add color function with arguments'
git add func_arg.py
git func_arg_dict.py 

git commit - m 'add mailroom revised version & mailroon test'
git add mailroom.py
git add mailroom_unittest.py


students/heb2016/session06/func_arg.py
```diff
@@ -0,0 +1,32 @@
...
+print("combination of position & keyword: ")
+color("purple", link_color='red', back_color ='blue')
+
+
+print("tuple & dict: ")
+links={'link_color': 'chartreuse'}
+regular=('red', 'blue')
+color(*regular, **links)
+
+
```
PythonCHB: looks like you're getting it -- make sense?

students/heb2016/session06/func_arg_dict.py
```diff
@@ -0,0 +1,25 @@
...
+    print("\n")
+
+print("default: ")
+color() 
+
+print("generic parameters keyword: ")
+color_args= ('green', 'black')
+color_kwags = {"link_color": "red", "visited_color": "blue"}
+color(*color_args, **color_kwags)
+
```
PythonCHB: try this calling the function with different combinations -- what happens when you do this?

color('blue', *color_args)

or

color('blue', **color_kwargs)

etc...

students/heb2016/session06/mailroom.py
```diff
@@ -0,0 +1,100 @@
...
+
+def print_report(donorlist):
+    print("This will print a report")
+    row=0
+    while row<=len(donorlist)-1:
+            print_list =donorlist[row]
+            row_msg="{:s},{:d},{:d}".format(*donorlist[row])
+            print(row_msg)
+            row+=1
+    return()
```
PythonCHB: You don't want to call  return !

either return None -- or just leave it off.

However, if you want to be able to test this, have it return a string with the report in it -- then print the report separately.

students/heb2016/session06/mailroom_unittest.py
```diff
@@ -0,0 +1,11 @@
+
+import unittest
+from mailroom.py import input_name
+
+def test_system(self):
+        raise SystemExit(1)
+        assert input_name('MS')
+
```
PythonCHB: what are you trying to test here?

In any case, you don't generally want to raise an Exception in a test function.

PR #100, opened 2016-11-08
Merge branch 'master' of https://github.com/UWPCE-PythonCert/IntroPython2016

spencermcghin:
Merge branch 'master' of https://github.com/UWPCE-PythonCert/IntroPython2016


students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -1,82 +1,151 @@
 #!/usr/bin/env python3
 
-import numpy, collections
+import numpy
+import collections
```
PythonCHB: good idea -- I do put multiple imports on one line, but only when they are related:

```
import os, sys
```

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -1,82 +1,151 @@
...
+            check_donor(prompt_donor_input)
 
 
+# Check if donor input is in current donor list, verify input is string #
+
+
+def check_donor(prompt_donor_input):
+    if prompt_donor_input in donors.keys():
+        print_email(prompt_donor_input)
+    elif prompt_donor_input != str(prompt_donor_input):
```
PythonCHB: won't prompt_donor_input already be a string?

students/spencer_mcghin/session3/mailroom_new.py
```diff
@@ -1,82 +1,151 @@
...
+    results = collections.OrderedDict()
+    donor_dict = {"Donors": []}
+    totals_dict = {"Total $": []}
+    # loop through donors data set and perform aggregate functions #
+    for donor, donations in sorted(donors.items()):
+        donor_dict["Donors"].append(donor)
+        totals_dict["Total $"].append((sum(donations)))
+    # combine dictionary objects into one for tabulate data input format #
+    results.update(donor_dict)
+    results.update(totals_dict)
```
PythonCHB: results is an OrderedDict -- but you are putting unordered data into it -- so why use the OrderedDict?

or do you want the totals after the donors? but don't care about the order otherwise?

This _may_ have worked if you got lucky, but don't count on the order of a regular dict!

PR #101, opened 2016-11-12
coding bat TDD

Baumelbi:
added some test code I wrote


Examples/Session06/cigar_party.py
```diff
@@ -8,8 +8,13 @@
...
+<<<<<<< HEAD
+def cigar_party(cigars, weekend):
+    return cigars >= 40 and (cigars <= 60 or weekend)
+=======
 
 
 def cigar_party(num, weekend):
     return num >= 40  and (num <= 60 or weekend)
 
+>>>>>>> 82918d5472627037e6cfaa6846d007c78ac8ef6c
```
codedragon: You will notice that there are two functions here that are very similar, but slightly different. Somehow you accidentally changed files in the Examples directory, and then when it asked about merging, you merged without editing to choose which one you wanted to keep. What would have been best, is if when you did git status, you had noticed that files in the Examples directory had changed, and then did a ```git checkout Examples/Session06/cigar_party.py``` in order to get the original file back before committing anything. In the future, you should copy files from the Examples directory to your directory, before editing. At this point, could you move these two files to your directory, and then do this?:
```git checkout d7d4e6042811c03e3 Examples/Session06/cigar_party.py```
The weird letter/number thing is the commit id of the last commit Chris made. I got it from doing 'git log'. Do the same thing for test_codingbat.py (same commit id, just the file name different). I hope this makes sense. Let me know if you have questions. I will wait until you do this to do the merge request, so we don't pull your changes to Examples into the common code.

File: students/baumel/session_06/test_in1to10.py [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+def test_OutRange_True15():
+	assert in1to10(15, True) == True
+
+def test_OutRange_TrueLower():
+	assert in1to10(-15, True) == True
+
+def test_base3():
+	assert in1to10(11, True) == True
+
+
```
codedragon: good coverage. Important to include the edge cases. :)

Baumelbi:
I believe I fixed the common code, sorry about that. let me know if i didnt fix the error

codedragon:
Looks better, thanks. I didn't see anything truly new, so didn't look closely. If there is something you still wanted us to look at, please let us know. 
cheers,
Maria

PR #103, opened 2016-11-15
wk6 hw

BenBagBag:


students/cowhey/session06/mailroom_test.py
```diff
@@ -0,0 +1,36 @@
...
+    db = mr.donors
+    assert "Mahesha Diogenes" in db
+    assert "Luuk Sinclair" in db
+    assert db["Neoptolemus Yaropolk"] == [36.00]
+
+
+def test_report_printing():
+    row = ['Neoptolemus Yaropolk', '36.00', '1', '36.00']
+    assert mr.format_row(row) == "Neoptolemus Yaropolk           36.00                          1                              36.00"
+
```
codedragon: Another way to do this would be to count the spaces:
```
rendered = mr.format_row(row)
rendered.count(' ', 0, rendered.find('36'))
```
and then compare the number of spaces you expect between the first and second words. You can do this for the space between each column.

students/cowhey/session06/mailroom_test.py
```diff
@@ -0,0 +1,36 @@
...
+    assert os.path.isfile("Neoptolemus_Yaropolk.txt")
+    with open("Neoptolemus_Yaropolk.txt") as f:
+        file_size = len(f.read())
+    assert file_size > 0
+
+
+def test_total_donation_amount():
+    """test that correct donation amount is returned"""
+    row = ['Arthur Tjaz', '18.00', '1', '18.00']
+    assert mr.total_donation_amount(row) == 18.00
```
codedragon: Good start to testing. Think about the important things that are being done, such as calculating the average donation, and how to test these things. You may need to refactor your code in order to test some things.

students/cowhey/session06/mailroom_rewrite.py
```diff
@@ -0,0 +1,146 @@
...
+    donor_rows = create_reports()
+    donor_rows.sort(key=total_donation_amount, reverse=True)
+    print(format_row(header_row))
+    for row in donor_rows:
+        print(format_row(row))
+    print()
+
+
+def send_thanks():
+    print("This will write a thank you note")
```
codedragon: This is a long function. Seems like it is probably doing several things, and could probably be broken into more pieces. In general, when refactoring, try to separate logic from formatting data.

PR #104, opened 2016-11-16
Session06 - Advanced Function Call and Unit Tests

psbriant:
Added Kwargs lab, testing lab, refactored mailroom and mailroom tests.


students/psbriant/session03/mailroom.py
```diff
@@ -14,32 +14,58 @@
 # -------------------------------Functions--------------------------------------
 
 
-def print_report(donors):
+donor_dict = {'Bill Gates': [1000000, 2500000, 1800000], 'Sheryl Sandberg':
```
PythonCHB: minor point, but lining up stuff in a long literal like this makes it easeir to read and edit:

```
donor_dict = {'Bill Gates': [1000000, 2500000, 1800000],
             'Sheryl Sandberg': [200000, 100000, 130000],
             'Larry Page': [150000, 110000, 170000],
             ...

File: students/psbriant/session03/mailroom.py [hunk 2]
```diff
@@ -14,32 +14,58 @@
...
+
+To send a thank you: type "s"
+To print a report: type "p"
+To exit: type "x"
+"""
+
+
+def sort_key(item):
+    """
+    Takes in a list of ints and outputs the int at index one.
```
PythonCHB: does the docstring need updating? this returns the zeroth item of the zeroth item... looks like you want to grab the first character of the first string? I assume it works as expected, but good to keep your comments sn doctrings up to date!

students/psbriant/session03/mailroom.py
```diff
@@ -14,32 +14,58 @@
...
     # iterate through donors
     for i in range(len(total_list)):
         # Name, total donated, number of donations and average donation amount.
         print(total_list[i][1], total_list[i][0], total_list[i][2],
-              total_list[i][0] / total_list[i][2])
+              round(total_list[i][0] / total_list[i][2], 2))
 
 
-def sum_d(donors):
+def sum_d(donor_dict):
```
PythonCHB: this is a good idea -- it means that this function needs to know the details of the data structure, but the rest doesn't.

nice!

(can you see how, now that we know about classes, that ir might make sense to make these kinds of functions all class methods?)

students/psbriant/session03/mailroom.py
```diff
@@ -76,11 +102,9 @@ def main_menu_selection(message):
     return action
 
 
-def create_letter(donor):
+def create_letter(full_name, amount):
```
PythonCHB: also good -- now this function doesn't need to know how a donor object is represented.

students/psbriant/session03/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+Tests for mailroom
+"""
+
+# -------------------------------Imports----------------------------------------
+import mailroom
+
+# -------------------------------Functions--------------------------------------
+
+
+def test_list_donors():
```
PythonCHB: very nice.

students/psbriant/session03/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    assert donor is None
+
+
+def test_create_letter():
+    """
+    Test output of letter to ensure donor name and donation are present with
+    correct formatting.
+
+    Question:
+    Need to fix assert error when verifying start of letter.
```
PythonCHB: is there extra whitespace in there maybe?

I often do a .strip() before checking strings...

students/psbriant/session03/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+    assert "donation of $200.00!\n" in letter
+
+
+def test_print_report():
+    """
+    Test to ensure report is generated.
+    """
+    donor_db = mailroom.donor_dict
+    report = mailroom.print_report(donor_db)
+    print(report)
```
PythonCHB: so this will test that the function will run without errors, which is a good start. Can you add some checks on the results?

students/psbriant/session03/test_mailroom.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def test_sum_d():
+    """
+    Test to ensure donor statistics are generated.
+    """
+    donor_db = mailroom.donor_dict
+    stats = mailroom.sum_d(donor_db)
+    print(stats)
```
PythonCHB: same as above -- maybe test if the stats are correct -- at least a little bit. This would bass if sum_d returned nothing....

students/psbriant/session06/args_lab.py
```diff
@@ -0,0 +1,76 @@
...
+    and visited links of a website and display the them.
+    """
+    print("Foreground: {} Background: {}"
+          " Links: {} Visited"
+          " links: {}".format(*args))
+
+
+# ==============================================================================
+
+
```
PythonCHB: I think you get the idea.

students/psbriant/session06/test_codingbat_lab.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+# ==============================================================================
+
+
+def main():
+    """
+    Calls functions and produces output.
+    """
+
```
PythonCHB: this all looks great

PR #105, opened 2016-11-17
Week 7

jerbear4:


PythonCHB:
Do you have specific questions? we will try to help!

But I'll see what I see in your code, too.

students/jbearer/session07/ex40.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env/python3
```
PythonCHB: nice straightforward example -- does it make sense?

students/jbearer/session07/html_render/html_render.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self, content=None):
+        self.content = []
+        if content:
+            self.content.append(content)
+
+    def append(self, content):
+        self.content.append(content)
+
+    def render(self, out_file, ind=''):
+        out_file.write("<{}>\n".format(self.tag))
```
PythonCHB: exactly! nice.

students/jbearer/session07/html_render/test_html_render.py
```diff
@@ -0,0 +1,103 @@
...
+
+    e.render(outfile)
+
+    outfile.seek(0)
+    file_contents = outfile.read()
+
+    print(file_contents)
+
+    assert('This is some text') in file_contents
+    assert('This is some more text') in file_contents
```
PythonCHB: you should probably check that the <p> and </p> have been rendered correctly, too.

students/jbearer/session07/html_render/html_render.py
jerbear4: This was your code. I was trying to figure out how it worked.

PythonCHB: I thought I liked it :-) -- does it make sense?

PR #106, opened 2016-11-20
SM: Session4 and Session5 Classwork

sasimandava:
Session4 and Session5 Classwork!


PythonCHB:
This is looking good now -- sorry for the mess!

PR #107, opened 2016-11-20
Wk7

BenBagBag:


students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+class Element:
+    tag = "html"
+    def __init__(self, content="", **kwargs):
+        self.content = []
+        self.attributes = []
+        if content:
+            self.content.append(content)
+        for k in kwargs:
+            # think about using ordered dict
+            self.attributes.append((k, kwargs[k]))
```
PythonCHB: note that this is going to put the attributes in the list in the order that the dict provides them, so still not deterministic -- but they won't change after this, at least.

Also -- this could be a bit cleaner with:

```
for k, val in kwargs.items():
    self.attributes.append((k, val))

or even a direct list consruction: 
self.attributes = list(kwargs.items())

File: students/cowhey/session07/html_render.py [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+        self.content.append(content)
+
+
+    def render_opening_tag(self):
+        attr = ""
+        for x, y in self.attributes:
+            attr += " "
+            attr += "{}=\"{}\"".format(x, y)
+        return attr
+
```
PythonCHB: can you do this with a " ".join()

and maybe a list comp?

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def render(self, out_file, ind=""):
+        out_file.write("{}<{}{}>\n".format(ind, self.tag, self.render_opening_tag()))
```
PythonCHB: nice!

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+        for item in self.content:
+            try:
+                new_ind = ind + "    "
```
PythonCHB: if you make a "indent" class attribute, then you could make this be:

new_ind = ind + self.indent.

Then you could change it on one place, on the class attribute, and it would change everywhere.

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+                item.render(out_file, ind=new_ind)
+                out_file.write("\n")
+            except AttributeError:
```
PythonCHB: love that EAFTP !

You generally want to put the Except as close as you can to where it may be raised:

                item.render(out_file, ind=new_ind)
            except AttributeError:
                out_file.write(new_ind + str(item) + "\n")
            out_file.write("\n")

that way you can be more sure where the AttributeError came from. For instance, what if someone passed in a not-file-like object that didn't have a write() method?

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+    tag = "head"
+
+
+class Html(Element):
+    tag = "html"
+
+    def render(self, out_file, ind=""):
+        out_file.write("<!DOCTYPE html>\n")
+        m = Meta()
+        m.render(out_file, ind="")
```
PythonCHB: nice! I like this. note that you don't have to give the tag a name to use it, of you only need to use it once.

Meta().render(out_file, ind="")

and you may want to pass ind through -- usually the html tag won't be indented, but if the render method takes the argument, it should pass it through.

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+    tag = "li"
+
+
+class Ul(Element):
+    tag = "ul"
+
+
+class OneLineTag(Element):
+    def render(self, out_file, ind=""):
+        out_file.write("{}<{}{}>".format(ind, self.tag, self.render_opening_tag()) + str(self.content[0]) + "</{}>".format(self.tag))
```
PythonCHB: that line's a little long... ;-)

PR #108, opened 2016-11-21
session 7 homework - html render

darrylsw:
code works up to step 3 of the run_html_render test, fails on step 4.

### DIFF
--- a/students/darrylsw/session07/html_render.py
+++ b/students/darrylsw/session07/html_render.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+class Element:
+
+	tag = ""
+
+	def __init__(self, content=None):
+		self.content = []
+		if content:
+			self.content.append(content)
+
+	def append(self, content):
+		self.content.append(content)
+
+	def render(self, out_file, indent):
+		tag_out = indent + "<" + self.tag + ">\n"
+		out_file.write (tag_out)
+		for i in self.content:
+			if isinstance(i, Element):
+				indent2 = indent + indent
+				i.render(out_file, indent2)
+			else:
+				out_file.write(indent + i + "\n")
+		tag_out = indent + "</" + self.tag + ">\n"
+		out_file.write (tag_out)	
+
+class Html(Element):
+	tag = "html"
+
+	def render(self, out_file, indent):
+		tag_out = "<" + self.tag + ">\n"
+		out_file.write (tag_out)
+		for i in self.content:
+			if isinstance(i, Element):
+				i.render(out_file, indent)
+			else:
+				out_file.write(indent + i + "\n")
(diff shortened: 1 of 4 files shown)


students/darrylsw/session07/html_render.py
```diff
@@ -0,0 +1,51 @@
...
+		if content:
+			self.content.append(content)
+
+	def append(self, content):
+		self.content.append(content)
+
+	def render(self, out_file, indent):
+		tag_out = indent + "<" + self.tag + ">\n"
+		out_file.write (tag_out)
+		for i in self.content:
```
PythonCHB: i try to use a more meaningful name than "i" for items in a list -- unless it's an index or the like.

students/darrylsw/session07/html_render.py
```diff
@@ -0,0 +1,51 @@
...
+				i.render(out_file, indent2)
+			else:
+				out_file.write(indent + i + "\n")
+		tag_out = indent + "</" + self.tag + ">\n"
+		out_file.write (tag_out)	
+
+class Html(Element):
+	tag = "html"
+
+	def render(self, out_file, indent):
```
PythonCHB: you don't want to completely re-write a method -- is this even different than the Elements render() method? Even if it's a little different, you want to re-use as much code as possible.

students/darrylsw/session07/html_render.py
```diff
@@ -0,0 +1,51 @@
...
+	tag = "head"
+
+class Title(Element):
+	tag = "title"
+
+class Body(Element):
+	tag = "body"
+
+class P(Element):
+	tag = "p"
```
PythonCHB: this is all looking good.

students/darrylsw/session07/test_html_render.py
```diff
@@ -0,0 +1,128 @@
...
+	assert file_contents.startswith ("<html>\n")
+	assert file_contents.strip().endswith ("</html>")
+
+
+def test_body():
+	outfile = io.StringIO()
+	b = Body("this is some text")
+	b.append ("and some more text")
+
+	assert "and some more text" in b.content
```
PythonCHB: you don't want to keep testing the same things -- ideally each test checks something new.

students/darrylsw/session07/test_html_render.py
```diff
@@ -0,0 +1,128 @@
...
+	assert b.content == ["this is some text", "and some more text"]
+
+	b.render(outfile, "")
+	outfile.seek(0)
+	file_contents = outfile.read()
+
+	assert ("this is some text") in file_contents
+	assert ("and some more text") in file_contents
+
+	assert file_contents.startswith ("<body>\n")
```
PythonCHB: this is the new and relevant part.

PR #109, opened 2016-11-21
html render

charlesrobison:
Stuck on step 2...a little help?


students/crobison/session07/html_render.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
+
+class Element:
+    def __init__(self, content=None):
+        self.content = []
+        self.tag = "html"
```
PythonCHB: you want the tag attribute to be a class attribute:

```
class Element:
    tag = 'html'
    def __init__(.....
```
that way every instance will have the same tag -- which is what you want. And you can change that tag easily in one place in a subclass.

students/crobison/session07/html_render.py
```diff
@@ -0,0 +1,55 @@
...
+                out_file.write("\n")
+            except AttributeError:
+                out_file.write(new_ind + str(stuff) + "\n")
+        out_file.write("</{}>\n".format(self.tag))
+
+    # def render(self, out_file, indent=""):
+    #     out_file.write("<{}>\n".format(self.tag))
+    #     for stuff in self.content:
+    #         new_ind = indent + "    "
+    #         if self.tag == "p":
```
PythonCHB: I think you see you're headed in the wrong direction here -- this approach would be very hard to extend.

The wonder of subclassing and polymorphism is that you don't have to know what type of element you are working with (even with self!) all you need to know is that it has an appropriate method to do the right thing.

PythonCHB:
I'll merge this, so we'll know what's different the next time around.

PythonCHB:
One more note:

Write some tests!! You want to know that each piece is working as it should all by itself, before putting it all together...

PR #110, opened 2016-11-22
Session 07: WIP and backlog of other items

jmhthethird:
Apparently I've been forgetting to PR my work week to week. :/


PythonCHB:
Whoa! yup, that's a lot :-) 
-- I wont' review any of the old stuff -- but let me know if you have any questions about any of it.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -0,0 +1,64 @@
+#!/Users/jhefner/python_dev/uw_python/bin/python
+
+class TextWrapper: #lifted this code directly from solution, explanation at line 20
```
PythonCHB: fair enough -- but make sure you "get it".

students/jhefner/session07/html_render/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+class Element():
+
+    html_tag = 'html'
+    indent = '    '
+
+    def __init__(self, content=None):
+        self.content = []
+        if content:
+            self.append(content)
+            # i dont understand this portion and why we put "content" there
```
PythonCHB: a bit confusing, cause we used the same name, but:

"content" is what the user passed in.

"self.content" is the list that we are storing the pieces of content in.

So we need to put the piece that the user passed in into the list.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def append(self,content):
+        # taking this append code from the solution.
+        # the lab text says to reference duck typing and whatnot.
+        # after reading the solution code and using it here I get why we are
+        # doing it this way but im not quite sure how i was supposed to
+        # know to arrive at this solution or something similar.
+        # OO stuff is still very new to me so maybe its just a pattern
```
PythonCHB: well, that's why I put it in the notes in the Exercise -- this is pretty complex and subtle stuff.

This is what's known as "polymorphism" in some languages -- the idea is that you can have multiple types of objects that can be used in the same way. In this case, we are making an object that can be used like an Element -- i.e. it can render itself.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+        out_file.write(ind+"<"+self.html_tag+">")
+        for i in self.content:
+            i.render(out_file)
+        out_file.write("</"+self.html_tag+">")
+
+class Title(OneLineTag):
+    tag = 'title'
+
+class Html(Element):
+    html_tag = 'html' #isnt this redundant? Shouldn't I leave the base class with an empty string for html_tag?
```
PythonCHB: yes -- it's probably better for the base class to have an empty string, or dummy value ("tag", maybe?).

but it's also useful, particularly when getting started, to have something as a default so the base class can function by itself.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -0,0 +1,64 @@
...
+    indent = ''
+
+class Body(Element):
+    html_tag = 'body'
+    indent = '    '
+
+class P(Element):
+    html_tag = 'p'
+    indent = '        '
```
PythonCHB: you _may_ want to change the indent behaviour in some special case, but mostly you want it to be consitent.

why does a P element indent more?

and don't you want an html element to indent its contents?

students/jhefner/session07/html_render/test_html_output2.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+    <html>
```
PythonCHB: you usually don't indent the <html> tag after the <!DOCTYPE html> tag.

students/jhefner/session07/html_render/test_html_output5.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
PythonCHB: having said that -- I guess my examples did that -- oops!

the new solution I did this week does not :-)

students/jhefner/session07/html_render/test_html_render.py
```diff
@@ -0,0 +1,44 @@
+#!/Users/jhefner/python_dev/uw_python/bin/python
```
PythonCHB: time to add some of your own tests!

PythonCHB:
I'm going to merge so we can see what's new at the next PR.

PR #111, opened 2016-11-22
HTML Render

charlesrobison:
Still having trouble...


students/crobison/session07/html_render.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Element:
+    tag = "html"
+    indent = "    "
+
+    def __init__(self, content=None):
+        self.content = []
+        if content is not None:
+            self.content.append(content)
```
PythonCHB: this should be:

self.append(content) 

that way it will use this classes append() method, which will do the magic of applying the TextWrapper, etc.

students/crobison/session07/html_render.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def append(self, content):
+        if hasattr(content, 'render'):
+            self.content.append(content)
+        else:
+            self.content.append(TextWrapper(str(content)))
+
+    def render(self, out_file, ind=""):
```
PythonCHB: this all looks good.

PythonCHB:
you have a conflict in your html_render file -- not sure how that happened.

but try doing:

git pull upstream master

to get them cleaned up.

make a copy of your file first, just in case something goes awry.

PR #113, opened 2016-11-23
Session 7- HTML Renderer

psbriant:
Parts 1-3 are complete, 4 has tests but no code and code and tests are written for part 5 but have not been run.

### DIFF
--- a/students/psbriant/session07/html_render.py
+++ b/students/psbriant/session07/html_render.py
@@ -134,3 +134,11 @@ class A(OneLineTag):
     def __init__(self, link, content, **kwargs):
         kwargs['href'] = link
         Element.__init__(self, content, **kwargs)
+
+
+class Ul(Element):
+    tag = 'ul'
+
+
+class Li(Element):
+    tag = 'li'
--- a/students/psbriant/session07/html_render.py
+++ b/students/psbriant/session07/html_render.py
@@ -57,7 +57,7 @@ class Element:
         open_tag, close_tag = self.make_tags()
         file_out.write((ind * self.num_ind) + open_tag + "\n")
         for stuff in self.content:
-            stuff.render(file_out, (ind * self.num_ind) + self.indent)
+            stuff.render(file_out, (ind * self.num_ind))
             file_out.write("\n")
         file_out.write(ind + close_tag)
 
--- a/students/psbriant/session07/html_render.py
+++ b/students/psbriant/session07/html_render.py
@@ -115,7 +115,7 @@ class SelfClosingTag(Element):
         """
         Take in an output file and generate a self closing tag.
         """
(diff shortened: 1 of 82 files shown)


PythonCHB:
"4 has tests but no code" -- awesome!

students/psbriant/session07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def make_tags(self):
+        """
+        Generalizes the creation of tags and allows for tags to be created with
+        or without attributes.
+        """
+        attrs = " ".join(['{}="{}"'.format(key, val) for key, val in self.attributes.items()])
+        if attrs.strip():
+            open_tag = "<{} {}>".format(self.tag, attrs.strip())
+        else:
```
PythonCHB: if attrs is an empty string at this point, you can probably use the same format call to create the tag, rather than having this if -- putting an empty string in wil be a do-nothing operation.

students/psbriant/session07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+class Element:
+
+    tag = 'html'
+    indent = "    "
+
+    def __init__(self, content=None, **kwargs):
+        """
+        Initializes class attribute content and appends any value not equal to
+        none into and empty list.
```
PythonCHB: you want docstrings to describe what the class does, and how to use it, but not internal implementation.

students/psbriant/session07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+    def render(self, file_out, ind=""):
+        """
+        Take in an output file and generate a self closing tag.
+        """
+        file_out.write("{}<{} />".format((ind + self.indent), self.tag))
+
+
+class Hr(SelfClosingTag):
+    tag = 'hr'
+    indent = ""
```
PythonCHB: why reset indent here?

students/psbriant/session07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+class Br(SelfClosingTag):
+    tag = 'br'
+    indent = ''
+
+
+class A(OneLineTag):
+    tag = "a"
+
+    def __init__(self, link, content, **kwargs):
+        kwargs['href'] = link
```
PythonCHB: nice!

students/psbriant/session07/html_render.py
```diff
@@ -0,0 +1,152 @@
...
+
+class Li(Element):
+    tag = 'li'
+
+
+class H(OneLineTag):
+    tag = "H"
+
+    def __init__(self, size, content=None, **kwargs):
+        self.tag = "h" + str(int(size))
```
PythonCHB: no need to define the tag as a class attribute, and re-define it in the __init__. Though it doesn't hurt, either.

PR #114, opened 2016-11-23
Personal Project

jmhthethird:
Odd that a bunch of other commits are in this PR as well. These should have been merged in https://github.com/UWPCE-PythonCert/IntroPython2016/pull/110 a day ago.


PythonCHB:
I think all those commits show up because this is the same branch as your other PRs.

But it's showing only one file changed -- so we're good.

PythonCHB:
And the project looks great.

If you get a fraction of that functionality, you are good to go for this class.

Be sure to check out the build-in json module for generating the json.

-CHB

PR #115, opened 2016-11-26
circle.py

charlesrobison:
Getting error in test, can't figure out why.

### DIFF
--- a/students/crobison/session08/circle.py
+++ b/students/crobison/session08/circle.py
@@ -4,6 +4,8 @@
 # 2016.11.26
 # Circle Lab
 
+import math
+
 class Circle:
     def __init__(self, radius):
         self.radius = radius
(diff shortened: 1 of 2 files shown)


students/crobison/session08/circle.py
```diff
@@ -0,0 +1,27 @@
...
+    def diameter(self, val):
+        self.radius = val / 2
+
+    @property
+    def area(self):
+        return math.pi * self.radius ** 2
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        self = cls()
```
PythonCHB: In this case, "cls" is going to be the Circle class -- and it needs the radius passed in.

Is this the failing test?

-CHB

PythonCHB:
We didn't get that far on this in class -- so make sure you get the html render Excercise down before going much farther on this.

PR #116, opened 2016-11-27
FOR REVIEW: html_render.py

pvosper:
Apologies - I couldn't figure out how to resolve merge these conflicts; html_render.py was edited both by me and CHB during the class session. After getting the changes from the code review in class, I could still commit my working version of the file but the conflict remains.


students/pvosper/session07/test_html_render.py
```diff
@@ -0,0 +1,89 @@
...
+    assert file_contents.strip().endswith('</html>')
+
+def test_link():
+    outfile = io.StringIO()
+    a = A("http://google.com", "link")
+
+    a.render(outfile)
+    
+    outfile.seek(0)
+    file_contents = outfile.read()
```
PythonCHB: I didn't remember in class the first time, but StringIO objects have a .getvalue() method that returns the full contents of the "file" without needing to call seek(0)

students/pvosper/session07/test_html_render.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    print(file_contents)
+
+    assert file_contents.startswith('<a href="')
+    assert file_contents.strip().endswith('</a>')
```
PythonCHB: you probably want to make sure the url got in there.

PythonCHB:
The way git deals with conflicts is to put it all in your lap -- which it kinda has to do, as it can't know what you want.

If you do a merge and two different people edited the same line(s) of code, git has no way to know which version you want. so what it does is mark the file as conflicted, and puts BOTH versions of those lines in the file, marked with >>>>>>>>>> and <<<<<<<<<<<<

so you have code that is no longer valid python.

YOu need to go in and edit those files, and remove the conflict markers and the code you dont want, and leave the code you do.

then you git commit after you've made the changes (and you code now runs and passes tests...).

after that commit, git will know you've resolved the conflict.

you may need to do a:

git pull upstream master

to get my version into yours first, if you haven't already done that.

PythonCHB:
conflicts fixed -- nice job.

PR #117, opened 2016-11-28
SM: Week6 Homework and lab exercises

sasimandava:


students/smandava/session6/sm_args_kwargs_lab.py
```diff
@@ -0,0 +1,41 @@
...
+    return "\n".join(colors)
+
+
+    # test input1: color_args_kwargs('yellow', fore_color = 'green', back_color= 'orange', link_color='white', visited_color = 'red')
+    # test input2: color_args_kwargs(fore_color = 'green', back_color= 'orange', link_color='white', visited_color = 'red')
+    # test input3: color_args_kwargs().
+    # test input4: color_args_kwargs('yellow', link_color='white', visited_color = 'red')
+    # test input5: color_args_kwargs('yellow', other_color='silver', visited_color = 'red')
+    # test input6: color_args_kwargs('yellow', fore_color = 'green', back_color= 'orange', link_color='white', visited_color = 'red', other_color = 'silver')
+    # test input7: color_args_kwargs('yellow', fore_color = 'green', fore_color = 'orange')
```
PythonCHB: looks like you've got *args and **kwargs down -- which was the goal :-)

students/smandava/session6/sm_mailroom_refactor.py
```diff
@@ -0,0 +1,86 @@
...
+
+    donor_summary_dict = {donorname: [sum(donor_dict[donorname]), len(donor_dict[donorname]), (sum(donor_dict[donorname]) / len(donor_dict[donorname]))] for donorname in donor_dict.keys()}
+    sorted_donors = sorted(donor_summary_dict.items(),
+                           key=operator.itemgetter(1), reverse=True)
+    for donor in (sorted_donors):
+        donor_report.append(donor[0] + ' ' + str(donor[1][0]) + ' ' + str(donor[1][1]) + ' ' + str(donor[1][2]))
+    return "\n".join(donor_report)
+
+
+def send_thanks(user_input, d_amount=None):
```
PythonCHB: nice -- this is nice and testable.

students/smandava/session6/sm_mailroom_refactor.py
```diff
@@ -0,0 +1,86 @@
...
+            d_amount = int(d_amount)
+        except ValueError as error:
+            return (type(error).__name__)
+        donor_dict.setdefault(user_input, []).append(d_amount)
+        return (user_input, donor_dict.get(user_input))
+    else:
+        try:
+            d_amount = int(d_amount)
+        except ValueError as error:
+            return (type(error).__name__)
```
PythonCHB: why return the error like this?

what you want is to re-prompt the user for a valid number. If you aren't going to do that in this function, then you can let the exception be raised, and catch it where it is called.

students/smandava/session6/test_codingbat.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+# Test cases for string_times.
+
+def test_string_times_2():
+    assert string_times('python', 2) == 'pythonpython'
+
+
+def test_string_times_3():
+    assert string_times('python', 3) != 'pythonpython'
```
PythonCHB: looks like you've got the idea with unit testing.

students/smandava/session6/test_mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+    print(result)
+    assert 'Alfred Hitchcock' in result
+    assert 'Martin Scorsese' in result
+    assert 'Ridley Scott' in result
+    assert 'Steven Spielberg' in result
+
+def test_sendthanks_add_donation():
+    result = send_thanks('Steven Spielberg', 60)
+    print(result)
+    assert ('Steven Spielberg', [40, 20, 30, 10, 60]) == result
```
PythonCHB: be a bit careful about hard-coding this particular data in your tests. maybe:

```
assert result[1][-1] == 60
```

that would check that the 60 was added to this record's list of donations.

students/smandava/session6/test_mailroom.py
```diff
@@ -0,0 +1,57 @@
...
+    result = send_thanks('Sasi Mandava', 'abc')
+    print(result)
+    assert 'ValueError' == result
+
+def test_writeletter():
+    result = write_letter()
+    assert "Created Thank you letter!" == result
+
+def test_writeletter_exists():
+    assert os.path.isfile('/Users/Mandava/Documents/python/IntroPython2016/students/smandava/session6/thankyouletter_Steven Spielberg.txt') is True
```
PythonCHB: no need to assert somethign is True, you can simple do:

assert something.

it will pass it True and fail if not.

PythonCHB: you should call the code that writes the file in the same test as you test for the file -- your tests should not be required to be run in any particular order.

students/smandava/session6/test_sm_args_kwargs_lab.py
```diff
@@ -0,0 +1,65 @@
...
+    print(result)
+    assert 'green' in result
+    assert 'orange' in result
+
+
+def test_color_empty():
+    result = color_args_kwargs_t()
+    print(result)
+    assert ' ' in result
+    assert ' ' in result
```
PythonCHB: this is going to check for a space in the result -- but almost any result is likely to have a space. Maybe you want:

`assert results == ' '`

or:

`assert result.strip() == ""`

students/smandava/session6/test_sm_args_kwargs_lab.py
```diff
@@ -0,0 +1,65 @@
...
+    assert 'yellow' in result
+    assert 'Key not accepted' in result
+
+
+def test_color_onearg_fivekwargs_othercolor():
+    result = color_args_kwargs_t('yellow', fore_color='green', back_color='orange', link_color='white', visited_color='red', other_color='silver')
+    print(result)
+    assert 'yellow' in result
+    assert 'Key not accepted' in result
+
```
PythonCHB: you're getting the idea. this kind of thing is pretty hard to test!

PR #118, opened 2016-11-28
Wk8

BenBagBag:


students/cowhey/session08/sparse_array.py
```diff
@@ -0,0 +1,66 @@
...
+            start_ind = key_list.index(start) + 1
+            del self.sa[start]
+        else:
+            for x in key_list:
+                if x > start:
+                    start_ind = key_list.index(x)
+                    break
+        for x in key_list[start_ind:]:
+            self.sa[x - 1] = self.sa[x]
+        del self.sa[key_list[-1]]
```
codedragon: Not sure what will happen if someone deletes the first index and it is a zero. You should have tests specifically for this function, as it is important and an easy point of failure.

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+class Html(Element):
+    tag = "html"
+
+    def render(self, out_file, ind=""):
+        out_file.write("<!DOCTYPE html>\n")
+        m = Meta()
+        m.render(out_file, ind="")
+        out_file.write("\n")
+        super().render(out_file, ind)
+
```
codedragon: Fyi, <meta charset="UTF-8" /> is getting put in the file twice, once where it belongs and once under <!DOCTYPE html>. I find render_opening_tag a confusing name; what about render_attributes or something like that? A little nit-picky, but makes it more obvious what it is doing.

students/cowhey/session07/html_render.py
```diff
@@ -0,0 +1,112 @@
...
+
+class Ul(Element):
+    tag = "ul"
+
+
+class OneLineTag(Element):
+    def render(self, out_file, ind=""):
+        out_file.write("{}<{}{}>".format(ind, self.tag, self.render_opening_tag()) + str(self.content[0]) + "</{}>".format(self.tag))
+
```
BenBagBag: Er, no, it's called `render_opening_tag` because only the opening tag can have attributes, e.g.:
`<p class="some-class">A paragraph</p>`
^is valid, but:
`<p>A paragraph</p class="some-class">`
is not.
`render_opening_tag` is also used in the opening tag of elements with elements inside of them (see line 28). You're right that it's badly named, it should probably be something like `render_attributes`.

students/cowhey/session07/points.py
```diff
@@ -0,0 +1,23 @@
...
+print(p.x)
+
+
+class PointB(Point):
+    def __init__(self, name):
+        self.name = name
+        Point.__init__(self)
+
+p2 = PointB(3, 3, "Bob")
+print(p2.color)
```
BenBagBag: Ah. That makes sense.

students/cowhey/session07/test_html_render.py
```diff
@@ -0,0 +1,166 @@
...
+    b = Body()
+    p = P("a paragraph")
+    b.append(p)
+    h.append(b)
+    output = get_output(h)
+    output_lines = get_output_lines(h)
+    print(output)
+    assert output_lines[0] == "<!DOCTYPE html>"
+    assert output_lines[1] == "<meta charset=\"UTF-8\" />"
+
```
BenBagBag: Whoops, good call.

students/cowhey/session08/sparse_array.py
```diff
@@ -0,0 +1,66 @@
...
+            return IndexError
+
+    def __setitem__(self, key, value):
+        if value != 0:
+            self.sa[key] = value
+        elif key in self.sa:
+            del self.sa[key]
+        if key >= self._len:
+            self._len = key + 1
+
```
BenBagBag: Good call. I'll fix that in this week's PR.

students/cowhey/session08/sparse_array.py
BenBagBag: Good point.

PR #120, opened 2016-11-29
Session08

jmhthethird:
Getting PR going to track work as it progresses.


jmhthethird:
I'm still unsure why all these previous commits show up in this PR (same as my last PR).

My fork is completely up to date with upstream/master (this repo).

screenshot 2016-11-28 18 40 30

jmhthethird:
I would dig some looking at of the first draft of my personal project. I started coding it a little bit at a time while the course was happening before I knew we were doing a big project at the end.

To that point there is a lot of older bad habits in there that need to be refactored. But one of the bigger things in it that I was curious about was if ya'll thought there was opportunity for use of classes in it rather than just the function driven flow I have.

I'm still new to classes and wouldn't really consider it a "tool" I can use yet. So I basically had built this a little bit at a time using functions for everything.

Thanks!

PythonCHB:
since you are always doing your 
PRs from the same branch (your master) gitHub is keeping all the commits in the list, even though they have been merged.

I suppose we should have you make a new branch for each Excercise, but we wanted to not burry you in to much git from the start.

We can see the changed files since the last merge, which is what we really need to know.

-CHB

students/jhefner/personal_project/requirements.txt
```diff
@@ -0,0 +1,15 @@
+ansible==2.2.0.0
```
PythonCHB: excellent idea to create this file and ensure your environment is consistent.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -1,64 +1,100 @@
...
-            # i dont understand this portion and why we put "content" there
 
     def append(self,content):
-        # taking this append code from the solution.
-        # the lab text says to reference duck typing and whatnot.
-        # after reading the solution code and using it here I get why we are
-        # doing it this way but im not quite sure how i was supposed to
-        # know to arrive at this solution or something similar.
-        # OO stuff is still very new to me so maybe its just a pattern
-        # that im not used to here.
```
PythonCHB: well, this is less an OO pattern than a Python duck-typing pattern. In a more statically types language, you would need to have a very simpel Element base calss, that might be a single piece of text.

Also, the API would be simpler, so instaed of allowing users to append either a Element  object or a plain string, you'd make the do this:

h = Html(TextWrapper("this is some plain text"))

that next bit of code is simply a way to provide a simpler API -- let the code create the text wrapper if it's needed.

students/jhefner/session07/html_render/html_render.py
```diff
@@ -1,64 +1,100 @@
...
+class TextWrapper:
     def __init__(self, text):
         self.text = text
 
     def render(self, file_out, current_ind=""):
         file_out.write(current_ind + self.text)
 
 class Element():
-
     html_tag = 'html'
```
PythonCHB: re your note later -- yes, you could have Element have an empty tag, as it is a generic, tag-less Element that needs specialization before it's useful.

but then it's harder to test the code, as it doesn't do anything, so I like to have a base class like this have some reasonable "real" values for everything.

PR #121, opened 2016-11-29
Older assignments for week 5 and 6

atruong012:
Catch up assignments and labs.


PythonCHB:
As this is older stuff, I'm jsut merging it -- but let me know if you do have specific questions.

PR #122, opened 2016-11-29
html_renderer passes step 5

darrylsw:
added tag attributes and now able to pass step 5, hopefully able to work on the rest before class tonight.

### DIFF
--- a/students/darrylsw/session07/html_render.py
+++ b/students/darrylsw/session07/html_render.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+
+class Element:
+
+	tag = ""
+
+	def __init__(self, content=None):
+		self.content = []
+		if content:
+			self.content.append(content)
+
+	def append(self, content):
+		self.content.append(content)
+
+	def render(self, out_file, indent):
+		tag_out = indent + "<" + self.tag + ">\n"
+		out_file.write (tag_out)
+		for i in self.content:
+			if isinstance(i, Element):
+				indent2 = indent + indent
+				i.render(out_file, indent2)
+			else:
+				out_file.write(indent + i + "\n")
+		tag_out = indent + "</" + self.tag + ">\n"
+		out_file.write (tag_out)	
+
+class Html(Element):
+	tag = "html"
+
+	def render(self, out_file, indent):
+		tag_out = "<" + self.tag + ">\n"
+		out_file.write (tag_out)
+		for i in self.content:
+			if isinstance(i, Element):
+				i.render(out_file, indent)
+			else:
+				out_file.write(indent + i + "\n")
+		tag_out = "</" + self.tag + ">\n"
+		out_file.write (tag_out)
+
+class Head(Element):
+	tag = "head"
+
+class Title(Element):
+	tag = "title"
+
+class Body(Element):
+	tag = "body"
+
+class P(Element):
+	tag = "p"
--- a/students/darrylsw/session07/html_render.py
+++ b/students/darrylsw/session07/html_render.py
@@ -3,40 +3,50 @@
 class Element:
 
 	tag = ""
+	indent = "   "
 
-	def __init__(self, content=None):
+	def __init__(self, *content, **kwcontent):
 		self.content = []
+		self.attr= []
 		if content:
-			self.content.append(content)
+			for item in content:
+				self.content.append(item)
+			for key in kwcontent:
+				value = key + '="' + kwcontent[key] + '"'
+				self.attr.append (value)
 
-	def append(self, content):
-		self.content.append(content)
 
-	def render(self, out_file, indent):
-		tag_out = indent + "<" + self.tag + ">\n"
+	def append(self, *content, **kwcontent):
+		#self.content.append(content)
+		for item in content:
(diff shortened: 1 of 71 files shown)


students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
 class Element:
 
 	tag = ""
+	indent = "   "
 
-	def __init__(self, content=None):
+	def __init__(self, *content, **kwcontent):
 		self.content = []
+		self.attr= []
```
PythonCHB: why a list here? attributes are very naturally key:value pairs -- i.e a dict.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
 		if content:
-			self.content.append(content)
+			for item in content:
```
PythonCHB: I see you've dealt with teh tuple of content objects -- are you wanting to be able to pass in a whole bunch?

if so, then this will work -- though ou could use:

self.content.extend(content) 

and save a loop.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
+				self.content.append(item)
+			for key in kwcontent:
+				value = key + '="' + kwcontent[key] + '"'
+				self.attr.append (value)
```
PythonCHB: this works fine (which I'm sure you know) -- but I think it's better to keep the key:value structure of the attributes, and deal with the formatting at the render step.

maybe you'd want to do something with them later??

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
 
-	def append(self, content):
-		self.content.append(content)
 
-	def render(self, out_file, indent):
-		tag_out = indent + "<" + self.tag + ">\n"
+	def append(self, *content, **kwcontent):
```
PythonCHB: is content going to have attributes? Elements have two things: attributes and content -- content itself does not have attributes.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
+			self.content.append(item)
+		for key in kwcontent:
+			value = key + '="' + kwcontent[key] + '"'
+			self.attr.append(value)
+		
+
+	def render(self, out_file, ind=""):
+		#tag_indent=self.initial_indent # or initial_indent=indent
+		tag_out = ind + "<" + self.tag
+		for att in self.attr:
```
PythonCHB: see if yu can elimate this loop with a " ".join(...)

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
+			tag_out += " "
+			tag_out += att
+		tag_out += ">\n"
```
PythonCHB: and maybe use a .format() to put all this in in one step.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
 		out_file.write (tag_out)
-		for i in self.content:
-			if isinstance(i, Element):
-				indent2 = indent + indent
-				i.render(out_file, indent2)
+		for stuff in self.content:
+			if isinstance(stuff, Element):
+				# look into using if hasattr (i, 'render')
```
PythonCHB: yup, checking for an attribute is more flexible.

or try calling stuff.render(), and catch the AttributeError if it's not something that knows how to render itself.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
+				indent2 = self.indent + ind
+				stuff.render(out_file, indent2)
 			else:
-				out_file.write(indent + i + "\n")
-		tag_out = indent + "</" + self.tag + ">\n"
+				indent2 = self.indent + ind 
+				out_file.write(self.indent + ind + stuff + "\n")
```
PythonCHB: you've defined indent2 -- why not use it here.

students/darrylsw/session07/html_render.py
```diff
@@ -3,40 +3,50 @@
...
-		tag_out = "<" + self.tag + ">\n"
-		out_file.write (tag_out)
-		for i in self.content:
-			if isinstance(i, Element):
-				i.render(out_file, indent)
-			else:
-				out_file.write(indent + i + "\n")
-		tag_out = "</" + self.tag + ">\n"
-		out_file.write (tag_out)
+	indent = ""
```
PythonCHB: html could still use the indent attribute -- isn't it used to indent the content within the html tag?

students/darrylsw/session07/html_render.py
```diff
@@ -48,4 +58,7 @@ class Body(Element):
 	tag = "body"
 
 class P(Element):
-	tag = "p"
\ No newline at end of file
+	tag = "p"
+
+class Hr(Element):
```
PythonCHB: an hr tag is self-closing -- you'll want to subclass from SelfClosingTag -- which will have a different render() method.

PR #123, opened 2016-11-30
Circle lab w/unit tests for review

pvosper:
I resolved conflicts locally, but Pull Request still says 'Can’t automatically merge. Don’t worry, you can still create the pull request.'


pvosper:
Sweet! "This branch has no conflicts with the base branch" - worked

students/pvosper/session09/sparse_array.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+'''
+
+class SparseArray:
+    def __init__(self, seq):
+        self.length = len(seq)
+        self.sparse_dict = {}
+        for i in range(len(seq)):
+            if seq[i] > 0:
```
PythonCHB: you should use != -- you do want to store negative numbers.

though I don't think I tested for that! -- hard to figure out everything you want to test!

students/pvosper/session09/sparse_array.py
```diff
@@ -0,0 +1,54 @@
...
+It should take a sequence of values as an initializer:
+
+
+'''
+
+class SparseArray:
+    def __init__(self, seq):
+        self.length = len(seq)
+        self.sparse_dict = {}
+        for i in range(len(seq)):
```
PythonCHB: maybe enumerate here:

for i, val in enumerate(seq):
    ....

students/pvosper/session09/sparse_array.py
```diff
@@ -0,0 +1,54 @@
...
+            if seq[i] > 0:
+                self.sparse_dict[i] = seq[i]
+
+    def __len__(self):
+        return self.length
+        
+    def __getitem__(self, i):
```
PythonCHB: perfect!

students/pvosper/session09/sparse_array.py
```diff
@@ -0,0 +1,54 @@
...
+        if i > self.length - 1:
+            raise IndexError('list index out of range')
+        elif i in self.sparse_dict.keys():
+            return self.sparse_dict[i]
+        else:
+            return 0
+
+    # This is crazy but works
```
PythonCHB: it is an inherently ugly problem :-)

I think it could be a bit cleaner, but I'm not seeing how right now....

if it works, you're good to go.

note how inefficient it is -- having to move everything around when you delete.

students/pvosper/session09/test_sparse_array.py
```diff
@@ -0,0 +1,100 @@
...
+    sa = SparseArray([7,29,0,0,0,0,13,0,0,72])
+    assert len(sa) == 10
+    
+def test_index():
+    sa = SparseArray([7,29,0,0,0,0,13,0,0,72])
+    assert sa[1] == 29
+    assert sa[2] == 0
+    assert sa[9] == 72
+
+# This works (I think) but I can't figure out how to test it
```
PythonCHB: you need a utiltiy from pytest (or from another test runner):

import pytest

```
with pytest.raises(IndexError):
    val = sa[21]
```

that will pass if an indexError is raised -- fail if not.

PythonCHB:
Looks like no conflicts -- yeah!

PR #126, opened 2016-12-05
SM: Class7 HTML Renderer exercise

sasimandava:


students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+    def __init__(self, content="", **kwargs):
+        self.content = []
+        # print(kwargs)
+        if content.strip():
+            self.content.append(content)
+        self.kwargs = kwargs
+
+
+    def append(self, content=""):
+        self.content.append(content)
```
PythonCHB: careful, this could get you an empty string in content again.

I'd make that a required argument:

def append(self, content):
   ...

and/or check for an empty string.

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+    def render(self, html_out, ind=""):
+        # print("Rendering ind, self:", len(ind), self)
+        # print("html_out.getvalue:", repr(html_out.getvalue()))
+        html_out.write(ind)
+        # print("After ind:", repr(html_out.getvalue()))
+        attrs = "".join([' {}="{}"'.format(key, val) for key, val in self.kwargs.items()])
+        # print(self.kwargs)
+        html_out.write("<" + self.tag + attrs + ">""\n")
```
PythonCHB: maybe make this a format:

"<{}{}>\n".format(self.tag, attrs)

adding a bunch of strings is less efficient

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+
+class P(Element):
+    tag = "p"
+
+
+class Head(Element):
+    tag = "head"
+
+class OneLineTag(Element):
+    def __init__(self, content="", **kwargs):
```
PythonCHB: if this is the same as the Element __init__ -- no need to write it again,

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+class Hr(SelfClosingTag):
+    tag = 'hr'
+
+class Br(SelfClosingTag):
+    tag = 'br'
+
+class A(Element):
+    tag = 'a'
+    def __init__(self, link, content):
+        self.link = link
```
PythonCHB: why store this here?

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+        self.content = content
+        Element.__init__(self, content, href=self.link)
```
PythonCHB: and this -- if you pass content into the Element __init__ -- you don't need to store it in the instance.

in fact, that self.content will get overridden anyway.

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+class Ul(Element):
+    tag = 'Ul'
+
+class Li(Element):
+    tag = 'Li'
+
+class Header(OneLineTag):
+    tag = 'h'
+    def __init__(self, intv, content):
+        self.intv = intv
```
PythonCHB: no need to store this, either. once the tag is constructed, you don't need it again.

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+            self.content = content
+        self.tag = 'h' + str(intv)
+        OneLineTag.__init__(self, content)
+
+class Doctype(Element):
+    tag = 'DOCTYPE html'
+    # def render(self, content):
+    def render(self, html_out, ind=""):
+        # self.content = content
+        html_out.write("<" + "!" + self.tag)
```
PythonCHB: if you are writing a custom renderer -- no need ot use the tag attribute, just hard-code it into the render method.

students/smandava/session7/html_render_new.py
```diff
@@ -0,0 +1,177 @@
...
+        # html_out.write('html')
+        html_out.write(">" + "\n")
+        # Element.render(self, html_out)
+
+class Encoding(SelfClosingTag):
+    tag = 'meta'
```
PythonCHB: a meta tag can hold other info -- so it makes sense to have a Meta element, and you can then put encoding in it as an attribute.

students/smandava/session7/test_html_render.py
```diff
@@ -0,0 +1,184 @@
...
+
+def test_anchor():
+    outfile = io.StringIO()
+    a = A("http://google.com", "link to google")
+    a.render(outfile)
+    file_output = outfile.getvalue()
+    print(file_output)
+    assert "http://google.com" in file_output
+
+def test_empty_content():
```
PythonCHB: nice!

PR #132, opened 2016-12-07
personal_project: review please.

jmhthethird:
Missing testing :(

I attempted to refactor the code a few times but ran into issues because the program requires CLI arguments to be passed to do anything. py.test (as far as I could find) doesn't really have a way to send in CLI arguments (such as -v, --list or other ones I defined). This blocked me for quite a while. Hopefully I can get some help on this from you guys on the last day of class.


PythonCHB:
I think we did talk about this in class, but for testing CLI programs:

1) you can test all the logic individualy -- each function should be able to be run with little setup.

2) you argparse creates an object with the results of the arguments parsed out. So you can create a dummy object for testing. something like:

```
class DummyArgs:
    test = False
    another_arg = 45
    ...
```

then you can set args to this for your tests, and you can change the values as you want them for each test.

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+import os
+import argparse
+import pymysql
+from configparser import SafeConfigParser
+import json
+
+"""
+argparse items
+"""
+parser = argparse.ArgumentParser(description="MySQL based Ansible Dynamic Inventory")
```
PythonCHB: good job figuring out argparse -- it's a bit klunky -- but useful.

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+group.add_argument('--list', action='store_true', help='list all hosts')
+group.add_argument('--host', action='store', help='list hostvars')
+args = parser.parse_args()
+
+
+#Thought: should i restructure all the below functions and their actions into
+# a class? im still a bit confused when a class is more beneficial than
+# these functions. im much more comfortable with functions so when thinking
+# through this i just kinda relied on functions as i knew how theyd operate.
+#
```
PythonCHB: a class would make it easier/cleaner to share data.

but if there is only ever going to be one instance of the class, then using the module namespace to share data is fine, too.

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Create pymysql connection based off of db config file in the
+    same directory.
+    """
+    parser = SafeConfigParser()
+    if os.path.isfile("db.ini"):
+        # TODO: is there a better way to read these variables and loop through
+        # and set them all at once? maybe have a base config.ini file that
+        # tells the program where to read the location of all the other files?
+        parser.read("db.ini")
```
PythonCHB: yeah, that would work. 

or put the config in a json file and read it with the json module -- goes straight into a nice python dict.

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+        dbhost = parser.get('db', 'host')
+        dbuser = parser.get('db', 'user')
+        dbname = parser.get('db', 'db')
+        dbpass = parser.get('db', 'pass')
+        conn = pymysql.connect(
+            host=dbhost, user=dbuser, db=dbname, passwd=dbpass)
```
PythonCHB: youcould put the calls to parser.get() right in here -- that would clean it up a bit:

```
conn = pymysql.connect(host=parser.get('db', 'host'),
                                       user=parser.get('db', 'user'),
                                       db = parser.get('db', 'db'),
                                       passwd = parser.get('db', 'pass'),
                                       )
```

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+        # that the applications config is incorrect.
+    return conn
+
+
+def query_db(conn):
+    """
+    take db connection handler and query db based on query within dbquery.ini
+    file in same directory
+    """
+    c = conn.cursor()
```
PythonCHB: check and see if this is a context manager -- i.e. can you do:

```
with conn.cursor() as c:
    ....
    ...
```
and things will be cleaned up properly.

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+    if args.list:
+        parser = SafeConfigParser()
+        parser.read("dbquery.ini")
+        dbquery = parser.get('query', 'list')
+        c.execute(dbquery)
+    if args.host:
+        parser = SafeConfigParser()
```
PythonCHB: a bit too muchduplicated code here. How about somethign like:

```
query = 'list' if args.list else 'host'
parser = SafeConfigParser()
parser.read("dbquery.ini")
dbquery = parser.get('query', query)
c.execute(dbquery)

return c.fetchall()

File: students/jhefner/personal_project/mysql_ADI.py [hunk 7]
```diff
@@ -0,0 +1,211 @@
...
+    So as you can see the for loops will go through each cell of the mysql
+    results and build a list of hostnames who have that same data. So we
+    should end up with a dictionary where the keys are all the unique results
+    from the database from every column and the "hosts" list should be all
+    the hostnames who have that value.
+
+    The output of this dicionary will be json. Due to how the json module
+    works it was easist to create the dictionary structure as you saw above
+    so that json can take the dict and output a json tree exactly how Ansible
+    is expecting it.
```
PythonCHB: yup -- nice that json objects map so well to python dicts!

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+        print("Result:", "\n", result, "\n")
+        print("Result type: ", type(result))
+        print("len(result):", len(result))
+
+
+    data = {}
+    for row in result:
+        data.update({row[1]:{}})
+        # data[str(row[1])] = {}  # machines.customer_id
+        data[str(row[2])] = {}  # assets.category
```
PythonCHB:
```
for i in range(1,5):
    data[str(row[i])] = {}
```
or maybe even something like (untested)

```
data = { str(row[i]): {} for i in range(1,6)}
```

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+        data[str(row[3])] = {}  # data_center_cages.name as 'DC_Cage'
+        data[str(row[4])] = {}  # assets.model
+        data[str(row[5])] = {}  # assets.make
+        data[str(row[1])]['hosts'] = []
```
PythonCHB: and this all the same, so you could put that in the above:

```
data = {str(row[i]): {'hosts' : []} for i in range(1, 6)}
```

12 lines reduced to one ;-)

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+
+    dictlist = list(data.keys())
+
+    if args.verbose:
+        print("\nDict key list:", dictlist)
+        print("len(dictlist):", len(dictlist))
+
+    for x in dictlist:
+        for row in result:
+            if str(row[1]) == x:
```
PythonCHB: do you need all these "str" calls -- is the data not already a string?

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+                data[str(row[1])]['hosts'].append(row[0])
+            if str(row[2]) == x:
```
PythonCHB: this could all be a loop, too:

```
for in in range(1,5):
    if str(row[i]) == w:
        data[str(row[i])]['hosts'].append(row[0])
```

students/jhefner/personal_project/mysql_ADI.py
PythonCHB: if row has non string content, -- you can clean it all up at once:

```
row = [str(item) for item in row]
```

students/jhefner/personal_project/mysql_ADI.py
```diff
@@ -0,0 +1,211 @@
...
+    # _meta is a constructor for Ansible. It is required in the output
+    # so I manually created it here and intentionally left its result empty.
+    data['_meta'] = {}
+    data['_meta']['hostvars'] = {}
+    return data
+
+
+def main():
+    if args.test:
+        # Test output, bypasses DB connection and provides test set of
```
PythonCHB: nice!

PR #136, opened 2016-12-10
Final Project for Review

jerbear4:


codedragon:
Nice job. It would be good to see the variables, such as the dictionaries like dict_path_iops, as well as file_list and myDate sent in and out of the functions as variables as needed. This would make everything much easier to test, and tests would also be nice to see. :)

PR #137, opened 2016-12-11
Final project no key

BenBagBag:
@PythonCHB I'll send you the API key so you can run this yourself if you want.


students/cowhey/final-project/app.py
```diff
@@ -0,0 +1,27 @@
+from flask import Flask, render_template, request, url_for
+from weather_data import WeatherData
+
+app = Flask(__name__)
+
+
+@app.route('/')
+def hello_world():
+    data = WeatherData()
```
PythonCHB: what does the data object look like? you may be able to do something really tricky:

render_template("index.html", **vars(data))

vars() returns a dict with the attributes of the object as keys, and the values of the attributes as the values.

but it doesn't give you class attributes -- so may not work. but if it does, it's pretty slick.

students/cowhey/final-project/templates/faq.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
PythonCHB: interesting FAQ.... :-)

students/cowhey/final-project/test_app.py
```diff
@@ -0,0 +1,26 @@
+from weather_data import WeatherData
```
PythonCHB: these don't exactly look comprehensive -- but a good start!

students/cowhey/final-project/weather_data.py
```diff
@@ -0,0 +1,69 @@
...
+                         "shorts": False,
+                         "sweatshirt": False,
+                         "boots": False,
+                         "sandals": False,
+                         "warm hat": False,
+                         "sun hat": False }
+        self.weather_dict = {}
+        if location:
+            self.location = location
+            self.userProvidedLocation = True
```
PythonCHB: why do you need to know of the user provided it?

but if you do, maybe simply have a location attribute that is None if not provided. -- or the empty string you have as the default.

students/cowhey/final-project/weather_data.py
```diff
@@ -0,0 +1,69 @@
...
+            self.location = self.get_location()
+            self.userProvidedLocation = False
+        self.get_weather()
+        # but now they're not printing all clothes
+        self.determine_clothing()
+
+    def get_location(self):
+        url = "http://freegeoip.net/json/"
+        location = json.loads(ur.urlopen(url).read().decode("utf-8"))
+        return "{}, {}, {}".format(location["city"], location["region_name"], location["country_code"])
```
PythonCHB: so this looks like the location is a dict -- so in your __init__ -- the default should probably be None, rather than an empty string. An empty dict is not a good idea, as it is mutable, and all instances would share the same one.

students/cowhey/final-project/weather_data.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_weather(self):
+        local_weather = owm.weather_at_place(self.location)
+        if self.userProvidedLocation:
+            self.location = local_weather.get_location().get_name()
```
PythonCHB: it seems like you  could put this logic in the __init__ -- os that here it wouldn't matter where the location came from.

students/cowhey/final-project/weather_data.py
```diff
@@ -0,0 +1,69 @@
...
+        current_weather = local_weather.get_weather()
+        self.weather_dict["status"] = current_weather.get_detailed_status()
+        self.weather_dict["rain"] = current_weather.get_rain()
+        self.weather_dict["snow"] = current_weather.get_snow()
+        self.weather_dict["temp"] = current_weather.get_temperature(unit="fahrenheit")["temp"]
+        self.weather_dict["humidity"] = current_weather.get_humidity()
+        self.weather_dict["wind"] = current_weather.get_wind()["speed"]
+        self.weather_dict["clouds"] = current_weather.get_clouds()
```
PythonCHB: so you have a dict here -- maybe that's what should get returned to your views, so you can pass it right inot the template.

students/cowhey/final-project/weather_data.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def determine_clothing(self):
```
PythonCHB: this is cute -- nice!

students/cowhey/session09/generator_solution.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        self.sum += self.current
+        self.current += 1
+        return self.sum
+
+def intsum2():
```
PythonCHB: generators are pretty slick, eh?

PR #138, opened 2016-12-11
Final project - first commit

darrylsw:
First commit for final project, working on grouping the pandas data so it can be imported into influxdb.


students/darrylsw/final_project/db_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+"""
+takes a csv listing all the datastore information from our vmware clusters such as storage used, free, capacity and 
+vms on each, sum up data based upon cluster and types of storage and import into a time series database, Influxdb. 
+This was what I discussed in my lightning talk toward the beginning of the quarter. The data will be used by Grafana 
+for different types of graphs.
+
+Importing of data into the database isn't functioning yet. The script is using classes so it can "easily" be expanded
+to accommodate other types of data such as vm and host counts, grouping by site, operating system, hardware,...
```
PythonCHB: sounds like a good plan.

students/darrylsw/final_project/db_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+"""
+
+import pandas as pd
+import os
+
+from influxdb import DataFrameClient, InfluxDBClient
+
+class df:
+	"""
+	dataframe class for default data
```
PythonCHB: these are big indents -- I can't tell in gitHub -- but are these tabs? Not a good idea -- you want to set your editor to use 4 spaces to indent.

students/darrylsw/final_project/db_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+	"""
+	tag_team = "vpt"
+
+	def __init__(self, input_file):
+		pdata = pd.read_csv(input_file)
+		ctime = os.path.getctime(input_file)
+	
+		# special actions for storage data
+		if self.tag_object == "storage":
```
PythonCHB: looks like tag_object is not being set until you subclass -- better to have a None or something in the base class.

students/darrylsw/final_project/db_ingest.py
```diff
@@ -0,0 +1,148 @@
...
+
+				json_allocated_pct = create_json ("Allocated_pct", self.tag_team, self.tag_object, tag_dstype, cluster, int(row['Date']), row['Allocated_pct'])
+				json_list.append (json_allocated_pct)
+		return json_list
+
+class vm_df (df):
+	"""
+	future class for vm data
+	"""
+	tag_object = "vm"
```
PythonCHB: careful here -- you want to override methods and attributes to get different functionality, rather than setting a flag, and then switching on that flag.

students/darrylsw/final_project/test_ingest.py
```diff
@@ -0,0 +1,59 @@
...
+	ds_df = storage_df(filename)
+	assert "Capacity" in ds_df.vmfs_data
+	assert "1.4814" in ds_df.vmfs_data['Date'].head(1).to_string()
+
+def test_group_data():
+	ds_df = storage_df(filename)
+
+	assert "vpt" == ds_df.tag_team 
+	assert "storage" == ds_df.tag_object
+
```
PythonCHB: maybe these should be separate tests?

students/darrylsw/final_project/test_ingest.py
```diff
@@ -0,0 +1,59 @@
...
+	assert "4.0039" in ds_df.vmfs_data['Allocated_tb'].head(1).to_string()
+	assert "5.8593" in ds_df.vmfs_data['Used_tb'].head(1).to_string()
+	assert ".75" in ds_df.vmfs_data['Used_pct'].head(1).to_string()
+	assert ".51" in ds_df.vmfs_data['Allocated_pct'].head(1).to_string()
+
+	assert "2.441406" in ds_df.nfs_data['Capacity_tb'].head(1).to_string()
+	assert "2.246094" in ds_df.nfs_data['Allocated_tb'].head(1).to_string()
+	assert "0.488281" in ds_df.nfs_data['Used_tb'].head(1).to_string()
+	assert "0.2" in ds_df.nfs_data['Used_pct'].head(1).to_string()
+	assert "0.92" in ds_df.nfs_data['Allocated_pct'].head(1).to_string()
```
PythonCHB: are you sure you want `to_string()` here? rather than `str(something)`?

if it's a number and you want to compare to some (but not all) digits, you can use:

```
import math
math.isclose(x, y, tol = 1e-6)
```
that will compare two numbers within about 6 digits -- very useful for testing.

PR #139, opened 2016-12-11
Final Project for review

pvosper:


codedragon:
Very cool. I like the small commits, and you have done pretty well at keeping the functions pretty contained. It is pretty unavoidable to have a lot of loops when you are going through lines in files in directories.

PR #140, opened 2016-12-11
Final Push for Quarter 1 with Final Assignment

ninadnaik:
Assignment file: ninadn_finalasn.py
Requires file: gamescore.csv in the same folder


students/ninadn/class5/mailroom_class.py
```diff
@@ -0,0 +1,91 @@
+'''Mailroom.py using classes for the various donors'''
```
PythonCHB: great idea -- I was thinking of assigning this to the whole class -- maybe next year.

students/ninadn/class5/mailroom_class.py
```diff
@@ -0,0 +1,91 @@
...
+
+class donor_info(object):
+	"""This class helps us store the name and donation information for each donor. Previously,
+	we did so by using a complicated list, tuple, list combination (I also tried dictionaries
+	to no avail. By using classes, we can simplify this process greatly, especially since we
+	will store similar information for each donor. Specifically, I want to use lists within
+	classes and am interested to see how that works."""
+	_registry = []
```
PythonCHB: this is pretty slick.

However, I would probably create a class for the while database of donors -- then you could put a bunch of logic in there.

students/ninadn/class5/mailroom_class.py
```diff
@@ -0,0 +1,91 @@
...
+		Q  Exit
+
+		Selection: ''')
+	return action.strip()
+
+if __name__ == '__main__':
+'''It is interesting in this solution, to get user options in the main function. I had previously 
+used the main function only to invoke a separare user input function, but that was redundant. While 
+it is not good style to pile on all the code in a few functions, all logical code should stay within
+a function; no need to break stuff down into too many functions or classes either.'''
```
PythonCHB: yup - you've found a pretty good balance.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
+'''Final assignment. This is a program that reads the gamescore.csv file, which has game information regarding a football match. 
+The program has a number of pre-created statements. Based on the score, by quarter (and assuming 2 OT), the program will
+print the box score, game summary, and auto-generate a short report about the game. The program can identify blowouts, close 
+finishes, and low scoring games. The long term aim of this project is to create a comprehensive program to scrape the internet
+for additional match information and create reports that offer historical perspective, etc.'''
+
+import numpy as numpy
```
PythonCHB: no need to import as the same name!

common use:

import numpy as np

which is only so you can save typing three characters each time you use it...

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+import pandas as pd
+import matplotlib as plt
```
PythonCHB: and this is usually 👍 
from matplotlib import pyplot as plt.

give you just the plotting namespace.

though it looks like you aren't using it yet in this code...

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+def printBoxScore(away_team, home_team, away_team_score, home_team_score):
+	star = '*'
+	print ('\n\n\n\n{:^100}'.format("BOX SCORE"))
+	print (star*100)
+	print (star," "*96,star)
+	print ('{:<2} {:>30} {:>10} {:^10} {:<10} {:<30} {:>2}'.format(star,away_team, away_team_score," ",home_team_score, home_team,star))
+	print (star," "*96,star)
+	print (star*100,'\n')
+
+def printDetailScore(away_team, home_team, P1ATS, P1HTS, P2ATS, P2HTS, P3ATS, P3HTS, P4ATS, P4HTS,P5ATS, P5HTS, P6ATS, P6HTS, away_team_score, home_team_score):
```
PythonCHB: that's a lot of parameters!

maybe this should all be in a dict or class?

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	dash = "-"
+	sep = "|"
+	us = "_"
+	print ('\n{:^99}'.format("GAME SUMMARY"))
```
PythonCHB: you could probably do this as one big format string.

then pass in a dict with all the data.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	print (sep,dash*95,sep)
+	print ('{:<2} {:<30} {:^8} {:^8} {:^8} {:^8} {:^8} {:^8} {:^8} {:>2}'.format(sep, away_team, P1ATS, P2ATS, P3ATS, P4ATS, P5ATS, P6ATS, P6ATS, sep))
+	print (sep,dash*95,sep)
+	print ('{:<2} {:<30} {:^8} {:^8} {:^8} {:^8} {:^8} {:^8} {:^8} {:>2}'.format(sep, home_team, P1HTS, P2HTS, P3HTS, P4HTS, P5HTS, P6HTS, P6HTS, sep))
+	print (sep,us*95,sep)
+
+def returnWinner(away_team_score, home_team_score):
+	'''If away team wins, return 1. If home wins, return 0. If tie, return 2.'''
+	if away_team_score > home_team_score:
+		winner = int(1)
```
PythonCHB: no need for int(1) -- 1 is already an int!

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	1	One team maintains lead throughout
+	2	Lead changes
+	3	Blowout
+	4	Close finish
+	5	Shut out
+	6	Low scoring
+	7	Generic
+	'''
+	gametype = []
+	blowoutMargin = int(21) #Defining a blowout as a points differential of at least 21 points
```
PythonCHB: more unneeded int() calls -- int() is for converting a float or sting to an integer..

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+	P1DIFF = int(P1ATS) - int(P1HTS)
+	P2DIFF = int(P2ATS) - int(P2HTS)
+	P3DIFF = int(P3ATS) - int(P3HTS)
+	P4DIFF = int(P4ATS) - int(P4HTS)
+	P5DIFF = int(P5ATS) - int(P5HTS)
+	P6DIFF = int(P6ATS) - int(P6HTS)
+
+	if numpy.sign(P1DIFF) and numpy.sign(P2DIFF) and numpy.sign(P3DIFF) and numpy.sign(P4DIFF) and numpy.sign(P5DIFF) and numpy.sign(P6DIFF) == 1:
```
PythonCHB: are you sure this doing what you think?

numpy.sign returns -1 or 1 -- neither are Falsey, so numpy.sign(P1DIFF) and numpy.sign(P2DIFF) is always going to return the second value. unless it's a tie, in which case it'll return zero, which will be Falsey -- then the entire expression will be False.

you may want to look at numpy.diff -- that will show you the difference between one score an the next. then you can do numpy,any() or numpy.all() to check some other stats.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	if int(home_team_score) == 0 or int(away_team_score) == 0:
+		gametype.append(5)
+
+	if int(home_team_score) < LowScoringThreshold and int(away_team_score) < LowScoringThreshold:
+		gametype.append(6)
+
+	if len(gametype) == 0:
+		gametype.append(7)
+
+	return gametype
```
PythonCHB: maybe use strings for game type? easier to read that numbers...

you could have a dict that maps names to numbers or vice-versa.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	P3ATS = answerlist[16]
+	P4HTS = answerlist[17]
+	P4ATS = answerlist[18]
+	P5HTS = answerlist[19]
+	P5ATS = answerlist[20]
+	P6HTS = answerlist[21]
+	P6ATS = answerlist[22]
+
+	'''I will simplify the above later on by using dicts where the key is the team and the values 
+	is a list of scores ordered by period.'''
```
PythonCHB: good idea.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+	'''In this function, I am reading the csv file using the with open method, then
+	going through the file line by line and putting each line in a list called infolist. 
+	As I do this, I am stripping out the \n at the end of each line, which was introducing 
+	an extra space after each line. Infolist is therefore a list where each element is a line
+	from the csv file. From this info list, I am splitting the first line and checking the 
+	element, which should tell me what the sport is. If the sport is football, I will proceed
+	or else I will exit as other sports are not supported. I then invoke the organize_data 
+	function which will split the infolist list into two separate lists, one with the key (which 
+	I call questionlist) and the other with the values (which I call answerlist). '''
+	with open ('gamescore.csv', mode = 'r', encoding = 'utf-8') as a_file:
```
PythonCHB: utf-8 is the default -- though it's good to be explicit.

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+		#print(a_file.read())
+		infolist = []
+		for line in a_file:
+			line = line.replace('\n',"")
+			infolist.append(line)
```
PythonCHB: strip() s a handy way to do this:

`infolist = [ line.strip() for line in a_file]`

students/ninadn/ninadn_finalasn.py
```diff
@@ -0,0 +1,237 @@
...
+		#print (infolist)
+		sport = infolist[0].split()[1]
+		#print (sport)
+		masterlist = []
+		if sport == 'football' or sport == 'Football' or sport == 'FOOTBALL':
```
PythonCHB: try:

if sport.lower() -- "football"

PR #142, opened 2016-12-12
Final Project

psbriant:
Final project with basic matplotlib functionality.


codedragon:
Looks like a cool project. I like the frequent commits. I'm guessing there is pretty big variation in LA water use, depending on which zip code you look at.

PR #143, opened 2016-12-12
SM: Python Class Project

sasimandava:
Python Class project with test cases and output file.

-Sasi


students/smandava/project/carrier_transid_count_refactor.py
```diff
@@ -0,0 +1,81 @@
...
+[3] Search for the occurrences of transaction IDs from server1 and IP address1 in the server2's log data and get final count.
+[4] Search for the occurrences of transaction IDs from server1 and IP address2 in the server2's log data and get final count.
+[5] Verify the counts in step 3 and step4 match.
+[6] Use decorators or context managers.
+"""
+import time
+"""
+Decorator function to calculate the execution time.
+"""
+def timed_func(carrier_transid_count):
```
PythonCHB: You usually use a generic name, like "function" here -- as you may put this decorator on anything...

students/smandava/project/carrier_transid_count_refactor.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@timed_func
+def carrier_transid_count(trans_file, log_file):
+    """Function to count the transaction IDs from server1 in server 2's logfile."""
+    total_tids = 0
+    server1_count = 0
+    server2_count = 0
+    with open(trans_file) as trans_file, open(log_file, encoding="utf-8", errors="ignore") as log_file:
+        """Ignore the errors encountered with parsing of characters like ±."""
```
PythonCHB: this is a comment, rather than a docstring -- use # for that.

students/smandava/project/carrier_transid_count_refactor.py
```diff
@@ -0,0 +1,81 @@
...
+        logfile_split = log_file.read().splitlines()
```
PythonCHB: could be log_file.readlines()

students/smandava/project/carrier_transid_count_refactor.py
```diff
@@ -0,0 +1,81 @@
...
+        for tid in trans_file:
+            total_tids += 1
+            for log_entry in logfile_split:
+                log_entry.strip()
+                tid.strip()
+                if (tid.strip() in log_entry.strip() and ("66.94.10.196") in log_entry.strip()):
```
PythonCHB: no need for a strip() call for an "in" check.

students/smandava/project/carrier_transid_count_refactor.py
```diff
@@ -0,0 +1,81 @@
...
+                    server1_count += 1
+                if (tid.strip() in log_entry.strip() and ("66.94.27.196") in log_entry.strip()):
+                    server2_count += 1
+                else:
+                    continue
```
PythonCHB: I think these continue calls are not needed -- you at the end of the function, continue is what happens automatically.

PythonCHB:
I hope to see you in the Spring.

PR #145, opened 2016-12-12
G-Coder Personal Project

aphollis:
This program takes user input and generates simple G-Code programs for cutting sheet goods on a CNC router.


codedragon:
This looks like a lot of fun to play with. With a little refactoring, you could make the code testable, which would be very useful to you as you start to make more complicated cuts.

PR #146, opened 2016-12-13
final project

PythonCHB:


codedragon:
Nice, useful scripts. I got into python through sysadmin, so it is close to my heart. :) Would be nice to see some tests.

PR #147, opened 2016-12-31
Final Project

weidnem:
Chris,

I pulled together a little routine that will help me in my job. I have this requirement to isolate the start and end hour of certain routes in the Metro bus system and export them to an Excel file. Currently information is compiled manually, but this will automate the process.

The routine pulls from the raw CSV files that powers the schedule view on Metro's website and from that isolates the start and end hour of each route.

If I had a bit more time I'd insert some error checking to catch connection failures when connecting to Metro website, as well as some other input checks, but I'm a little limited on time.

Let me know if you'd like me to restructure this in any way.

Thanks


students/weidnem/final_project/RouteProcessor.py
```diff
@@ -0,0 +1,90 @@
...
+@author: Matthew Weidner
+
+Read select a CSV files from the Metro website
+Find the greatest and smallest time value in each file
+output greatest and least time to a file.
+
+"""
+
+# import libraries
+from urllib import request
```
PythonCHB: take a look at the third party library: requests

pip install requests.

It's nicer to use than urllib. Though you've got it working, which is good.

students/weidnem/final_project/RouteProcessor.py
```diff
@@ -0,0 +1,90 @@
...
+
+def get_route(route):
+    # open connection
+    response = request.urlopen("http://kingcounty.gov/~/media/depts/transportation/metro/data/schedules-sept-2016/a0{}.csv".format(route))
+    
+    # read and store contents
+    csv = response.read()
+    
+    # decode to a str
+    csvstr = str(bytes.decode(csv))
```
PythonCHB: This will use the default encoding -- apparently the right one, but you may want to specify the encoding to be sure.

students/weidnem/final_project/RouteProcessor.py
```diff
@@ -0,0 +1,90 @@
...
+    
+    # split str into rows
+    lines = csvstr.split("\r\n")
```
PythonCHB: if this is a CSV file, you may want to use the csv module to read it -- it will take care of some of the subtleties for you.

students/weidnem/final_project/RouteProcessor.py
```diff
@@ -0,0 +1,90 @@
...
+    
+    times = []
+    hours = []
+    # interrogate rows, elimiate first two and last line in file
+    for line in lines[2:-1]:
+        times.extend(line.split(",")[:-1]) #omit HTML in last column
```
PythonCHB: there's html is in the csv file? ouch! -- looks ike you've dealt with it though.

students/weidnem/final_project/RouteProcessor.py
```diff
@@ -0,0 +1,90 @@
...
+    
+    #clean list, isolate hours of service
+    i = 0
+    while i < len(times):
+        try:
+            hours.append(int(times[i][:times[i].find(':')])) #entries with valid times will have a ':'
+            i += 1
+        except:
```
PythonCHB: what exception are you catching? ValueError?? be explicit, or some other error could pass through.

Also, if you are looking for a colon -- probably better to jsut do:

if ....find(':') >=0:
    ...

PythonCHB:
All Good!

I'll let UW know to remove the incomplete.

I hope I see you in the third quarter.

-Chris

